PR #63, opened 2014-01-12
City Regions

frankieroberto:
This is an idea that’s been kicking around for a while, and has suffered from a few failed referendums – but it makes a lot of sense and would help England's other major cities catch up with London.


Floppy:
Going to close this one, as it's long dead, and the devolution text in the manifesto has moved on substantially. @frankieroberto, feel free to revisit the idea based on the current text though :)

PR #191, opened 2014-06-18
Added remit: "Reducing inequality" for central bank

cyanc:
This is up for discussion. It would create a deliberate confict of interest between restricting price inflation by lowering interest rates, and the unintended consequences of such actions - increasing asset inflation (e.g. property prices and share prices). Asset inflation is good for the wealthy because they hold assets, the poor don't hold assets so lose out - e.g. housing becomes more expensive. This is an example of "predistribution"; rather than allowing assets to rise exponentially and then taxing them, take action to restrict their rise in the first place.


Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/191), and results are set in the merge status checks below.

## Changes

PR #192, opened 2014-06-19
changed council tax to property tax

cyanc:
Council tax is rubbish in so many ways it's not worth listing them here!

The idea of this change is that it acts as a counterbalance to above inflation property price rises (which increase inequality by favouring asset owners, and rising faster than working people can earn/ save).

Methodology:
The average house price was £260k as of April 2014 - http://www.ons.gov.uk/ons/rel/hpi/house-price-index/april-2014/stb-april-2014.html#tab-House-Price-Index-UK-Summary

0.5% of the average house price is around £1300, which is about the same as the average council tax (Band D average way £1400 this year). 0.5% of a £1m house would be around £5k per year. Higher % levies could be applied to higher value properties (e.g. 1% over £1m, 2% over £2m - negating the need for a mansion tax).

If the value of your house rises you pay more, if it falls you pay less. The owner, rather then the occupier pays, this means Landlords would be liable to paying on rental properties (so would need to pass costs into the rent) but the tax would be levied whether or not the property is occupied - discouraging vacancy/ hoarding.

Reasoning:
It will act as a brake on above wage-inflation house price rises by taking money out of the economy during a house price boom, and cement housing booms as a bad thing in the public consciousness, as well as acting as a cushion and stimulus in the event of a house price crash.

It will also mean that if the public pay (via taxation) to improve an area, those benefiting will contribute to it on an ongoing basis. e.g. if building Crossrail results in a new station near your house providing a direct commuter route into the City and your house rises in value, you contribute. At the moment all tax payers (including renters) pay for improvements like Crossrail, and only the property-owner benefits from the resulting property price rises.

Problems:
Who gets this money? If the local council continue to get it, rich areas will still collect more tax than poor areas. However, if it goes to central government, then how do local councils receive their funding?

### DIFF
--- a/manifesto/housing.md
+++ b/manifesto/housing.md
@@ -24,7 +24,7 @@ With fewer people able to afford to buy, many are renting. We will carry out a p
 
 ## Council Tax
 
-Investigate a fairer system than the current 'band' system of council tax.
+Council tax will be discontinued and replaced with a 'Property tax', set at 0.5% of the property's value and payable by the owner of the property whether it is occupied or not. The property's value will be calculated by an initial valuation, and then rise or fall annually in line with the ONS regional house price index (HPI). 
 
 ## Housing Benefit
 


Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/192), and results are set in the merge status checks below.

## Changes

PR #197, opened 2014-07-13
Replace extradition with UK based trials

jt-nti:
I suspect this needs more thought (I'm guessing renegotiating extradition treaties is unlikely to be simple!) but I think the current system is unbalanced. The burden should be firmly on the prosecuting country, which has far larger resources at its disposal than the average defendant who, after all, is innocent until proven guilty.
Trying people in the UK seemed like a good starting point, especially given the nature of some crimes that have caused problems (committed in the UK, or where there are doubts about fair trials abroad, potential for onward extradition, etc.). The intention is, if the details can be made to work, to end up with a much quicker process (fewer fights against extradition) and one that doesn't unduly punish anyone before they are convicted.
Any thoughts?
Cheers, James

### DIFF
--- a/manifesto/crime.md
+++ b/manifesto/crime.md
@@ -30,4 +30,8 @@ Implement rigorous protection regimes for whistleblowers.
 
 A review will be carried out into the composition of juries, especially for complex trials.
 
+## Extradition
+
+The UK will no longer extradite anyone to stand trial abroad. Alleged offences may have been committed while defendants were in the UK, or only in another country for a short time, and facing trial in an unfamiliar country without access to a support network that would available at home is too close to punishment rather than process. Instead, provisions will be made for 'international courts', where defendants can be tried without leaving the country. These international courts would host trials by countries which previously had extradition treaties with the UK.
+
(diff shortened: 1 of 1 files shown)


jthub:
"I think the basic rationale is that we're defaulting to protecting residents of the UK until proved guilty, rather than handing them over to some other country where we can't do that any more." sounds like a reasonable summary.

Monitoring and assurances still seem too weak to me- there is still effectively a element of punishment in the process. For example, I was trying to remember a fairly recent case where someone who had been extradited to the US and was experiencing medical problems- I think it was Paul Dunham who, at the time, had still not been convicted, was on bail, could not work, had heart problems and could not have medical expenses covered by the UK.

Legal aid sounds great, but that could be a wider issue!

Perhaps the defendant remain in the UK except when actually required to appear in court?

There also seems to be a group with a related campaign (I should probably read it when I get the chance!) - http://www.friends-extradited.org/index.php

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/197), and results are set in the merge status checks below.

## Changes

Issue #56, opened 2014-01-10
Registering a party


Floppy:
We should perhaps register a party. We'd need a name, and while the project is currently OpenPolitics, I'm not sure that's a great party name. Any better suggestions?

PR #199, opened 2014-07-21
Rename site to "The OpenPolitics Network"

Floppy:
This is just an idea to let people try the rename I suggested in #56 on for size. I've just changed the main title, and removed the paragraph that talks about not being a party yet but aiming to have candidates, as it was a little outdated. I've explicitly _not_ stated that we _are_ a party called the OpenPolitics Network, that's undecided as yet.

screen shot 2014-07-21 at 21 12 51


Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/199), and results are set in the merge status checks below.

## Changes

PR #201, opened 2014-08-01
Expansion of clause

CharlieDelMonte:
Current clause does not cater for defence of UK or commonwealth land (i.e. Falklands) or humanitarian missions (evacuation: http://edition.cnn.com/WORLD/9708/24/montserrat/)

### DIFF
--- a/manifesto/military.md
+++ b/manifesto/military.md
@@ -14,7 +14,9 @@ The United Kingdom should take a role as international leader in pursuit of glob
 
 ## Role of UK Armed Forces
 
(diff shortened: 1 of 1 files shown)


Floppy:
Just going through old changes, and found this again, and realised we never revisited Charlie's new wording. I think I'm OK with it... :+1:

Any thoughts, @PaulJRobinson @philipjohn?

philipjohn:
I'm a :hand: for two reasons:

* removal of "only" from the first sentence makes the policy completely limp, because it turns the policy into a guideline, not something that ministers have to abide by - we may as well not bother.
* "the government should justify the decision to act" is exactly what every government does anyway - they will say what they think they need to in order to justify the decision, but that's just PR - the decision itself must be subject to some boundaries, hence the original text requiring that unilateral action be the justification for UK involvement.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/201), and results are set in the merge status checks below.

## Changes

PR #215, opened 2014-10-05
Purpose of the United Kingdom

frabcus:
I think *the* main problem with the UK right now is that nobody knows what it is for.

Back round about 1950, a post-war goal of economic wealth through technology, along with longterm peace was made. In many ways we have now achieved that (despite its flaws). What is our goal now?

This is my proposal.

### DIFF
--- a/manifesto/index.md
+++ b/manifesto/index.md
@@ -5,6 +5,19 @@ published: true
 
 # The OpenPolitics Manifesto
 
+## Purpose
+
+Without a purpose, a country will ebb and flow on the selfish whimsy of its people.
+
+The purpose of the United Kingdom is:
+
+1) For everyone in the world to fully receive all the rights universally declared [^1]
+
+2) For that situation to be indefinitely achieved
+
+3) To then spend our time fully understanding ourselves, the earth and the universe through the arts, science and exploration.
+
+
 ## Principles
 
 We believe that government should serve the people, and that 'the state', so often derided, is simply the way that citizens work together to enhance the public good.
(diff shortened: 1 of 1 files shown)


philipjohn:
There's some kind of human rights index, right? Something like that maybe, coupled with other similar indexes based on poverty, health, education?

sanjaypoyzer:
Are there any other countries which have a mission statement like this? Otherwise we're comparing the UK to a charity (which maybe is a good thing, but it should definitely be noted how unprecedented it is.)

frabcus:
Well, the purpose of the US was originally "the preservation of life, & liberty, & the pursuit of happiness".

philipjohn:
I'm not sure a state is *for* anything, or should be, actually. People increasingly identify in ways that don't honour borders, and the internet makes this easier. We have a much more globalised culture than ever before where states are becoming less and less important to people other than politicians.

tmtmtmtm:
> Well, the purpose of the US was originally "the preservation of life, & liberty, & the pursuit of happiness".

See also: “Liberté, égalité, fraternité”. This also reminds me of Bhutan introducing Gross National Happiness as its primary metric. 

Or, from another angle, the mission of setting up a worldwide caliphate…

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/215), and results are set in the merge status checks below.

## Changes

PR #221, opened 2014-10-06
Implement a robin hood tax

Floppy:


philipjohn:
@Floppy Mind if I tweak this, if you're struggling to find the time?

Floppy:
Oh, god, yes, go ahead. Or, I'll do it as soon as I get a minute or two.

philipjohn:
Done! It's a :+1: from me.

@PaulJRobinson Could you add your vote again please?

@mikera @tmtmtmtm Keen to get your votes now on this proposal

Floppy:
lovely

tmtmtmtm:
I'm still finding the wording a little confusing, as it's not completely clear whether the two sentences are meant to be saying essentially the same thing, or whether it's two overlapping, yet distinct, proposals.

There's also a slight ambiguity as to what it means to "join" the tax: is the proposal to implement it:

1. no matter what anyone else does
2. as long as at least *some* other EU states do
3. simultaneously with a Eurozone-wide implementation
4. only if the entire EU does?

I'd happily vote yes on 3 and 4; the first two I'm a little more uncertain about. 

I'd also be open to a wording that says we'd help work towards a unanimous implementation, but would consider any lesser plan carefully.

philipjohn:
Sounds like you have a far better handle on this than we do @tmtmtmtm :) So in the interests of collaboration, do you fancy drafting some wording that you think adequately does the job? We'll have a consensus then too :)

Floppy:
@tmtmtmtm I've had another crack at the wording in an attempt to answer your question. See what you think.

Floppy:
@mikera the EU FTT is apparently designed not to impact the "real" economy. Worth having a read of the wikipedia writeup, see if it addresses your concerns.

tmtmtmtm:
The new version looks good to me. :+1:

philipjohn:
:+1:

Floppy:
Sorry guys, still tweaking; added a point about global system. Can you vote again?

Floppy:
I'm done now :)

philipjohn:
No :p :+1:

tmtmtmtm:
I'm a little hesitant on the new additional wording, I'm afraid — at least without further elaboration. What's the idea here?

Floppy:
basically any system will work better the more widely it's implemented, so as well as joining the EU one, we'd want to push for a global deal.

tmtmtmtm:
It may just be the wording but "work to create a global system to prevent avoidance" seems quite strong. We've already seen how difficult it is to even get an EU-wide version agreed — what sorts of actions are you proposing to take against countries who don't implement such a tax?

Floppy:
Tried to improve the wording again around the global FTT regime. Basically I'm trying to say that we'd join the EU FTT system, but with the understanding that to work properly in a globalised economic system, the tax needs to be global.

tmtmtmtm:
I'm a strong believer in meddling in other countries' affairs as little as possible, and I'm distrustful of most attempts to create a global regime for anything. But I'd be OK with wording like "We would also work with international organisations to broaden such a regime beyond the EU".

Floppy:
Ah, that's better. Thanks, I was just looking for decent wording :)

Floppy:
done

philipjohn:
:+1:

tmtmtmtm:
:+1: from me then too :)

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/221), and results are set in the merge status checks below.

## Changes

PR #230, opened 2014-10-07
Introduce overseas constituencies

tmtmtmtm:
Currently British citizens who move abroad still vote in Parliamentary elections in the constituency in which they were last registered, even though they often have no remaining ties to that location. Instead, introduce one or more overseas constituencies whose MP(s) specifically represent the people who now live there, similarly to the [French system](https://en.wikipedia.org/wiki/Constituencies_for_French_residents_overseas). This is even more important in the UK where there is a much closer tie between an MP and their constituents.

### DIFF
--- a/manifesto/democracy.md
+++ b/manifesto/democracy.md
@@ -21,6 +21,8 @@ Introduce a 'None of the Above' box on all ballot papers to formally and positiv
 
 Introduce voting by proportional representation using the Single Transferable Vote (STV) method in all General and Local Elections.
 
+Introduce new constituencies for British citizens living abroad. 
+
 Investigate the feasibility of electronic voting at all elections, with a view to increasing accessibility and turnout.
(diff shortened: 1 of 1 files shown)


Floppy:
@frankieroberto are you still opposed to this addition? There would be more detail to work out, but from the comments above, it doesn't seem you're opposed to the general principle. Any more recent thoughts? Obviously this will evolve further along with other constituency discussions like #219, but it would be great to get the general principle agreed.

Floppy:
Closing due to block, sadly.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/230), and results are set in the merge status checks below.

## Changes

PR #236, opened 2014-10-12
Parishes and CCs as the seat of LA policy making

digitalWestie:
I'm a fan of confederalism (http://www.inclusivedemocracy.org/dn/vol1/bookchin_confederalism.htm), perhaps this is how it could start off in the UK ?

### DIFF
--- a/manifesto/local_government.md
+++ b/manifesto/local_government.md
@@ -11,3 +11,7 @@ How do we improve and reinvigorate local government?
 ## Transparency
 
 Publish full minutes (or transcripts if available), audio recordings, and voting records from all local council meetings. Online publication should be performed within 24 hours. Software tools will be created to make this process simple.
+
+## Local policy making
+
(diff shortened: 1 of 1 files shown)


philipjohn:
Could you expand a bit on how this would work in practice? I'm not entirely understanding it at the moment.

Floppy:
Same here, I don't quite understand what would actually happen (though I like the idea of pushing policymaking out to the edges, obviously)

Floppy:
:hand: for now.

Note this is also over the time limit, so unless you do want to push it forward @digitalWestie, it will probably be closed soonish.

digitalWestie:
Yep still chewing this one over, since proposing I've found more sources from which to develop this further so will put something down eventually. Again this might become multiple proposals so will close for now. NB If you want to know what I'm on about, I'm aiming at a proposal that comprises some of the ideas / practices such as libertarian municipalism, citizen juries, citizen assemblies, and participatory budgeting.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/236), and results are set in the merge status checks below.

## Changes

PR #241, opened 2014-10-14
Invest in low emission vehicles

BenTrimble:
Changes made to try to make the pledge more in line with achievable end result. The existing technology available does not allow us to eliminate fossil fuels and hence we should look at the steps needed to achieve that first, with a commitment to extend into zero-emissions later.

### DIFF
--- a/manifesto/transport.md
+++ b/manifesto/transport.md
@@ -11,7 +11,7 @@ What policies could improve public and private transport infrastructure?
 
 ## Removal of Fossil Fuels
 
-We will decrease emissions from road transport by setting a date beyond which no new fossil-fuel powered vehicles may be sold in the UK.
+We will decrease emissions from road transport by ensuring that the technology to create low / zero emission transportation and related infrastructure is as cheap and freely available as possible. We will aim to set a date beyond which no new solely fossil-fuel powered vehicles may be sold in the UK and will pursue an agenda to eliminate fossil fuels from the transportation network entirely.
 
 ## National Infrastructure Organisations
 


philipjohn:
@Floppy Are you happy with this other than having a set date? If so, does it makes sense to :+1: this one and do another PR afterwards?

@BenTrimble Can you click the big button on https://www.clahub.com/agreements/openpolitics/manifesto please so that we can include your contributions?

Floppy:
Yeah, OK. :+1:

Floppy:
Since this was submitted, we've got a date in there for no new fossil fueled vehicles after 2029. This PR should, if we still want it, be updated to reflect that wording, or change it. As the source text has changed, it might be best for @BenTrimble to resubmit this change based on the new words maybe? I suspect he wouldn't be happy with just adding "by 2030" on the end, given this thrust of this change is to remove such talk of deadlines.

Floppy:
I'm moving to :hand: on this, as I think it probably needs redoing based on the current text.

Floppy:
We need @BenTrimble's input on this one, but seeing as the underlying text has changed substantially and this has passed the 90-day merge deadline, I'm going to close this and invite @BenTrimble to propose the update again based on the existing text if he wants to.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/241), and results are set in the merge status checks below.

## Changes

PR #248, opened 2014-11-12
Support worker owned enterprise

digitalWestie:
Some relevant links:
- http://employeeownership.co.uk/
- http://en.wikipedia.org/wiki/Worker_cooperative
- http://www.radicalroutes.org.uk/publicdownloads/setupaworkerscoop-lowres.pdf


philipjohn:
My understanding is that co-operatives mostly use the limited by guarantee form of limited company. Is this not suitable?

Floppy:
@digitalWestie how do you mean "we will introduce legislation which is designed with these types of organisation in mind."? Would that be better phrased as "we will create a new corporate structure explicitly designed for these organisations"?

A couple of questions, if that's the case:

1) As Phil says, what do the current structures not allow that co-operatives need? Limited-by-guarantee, CICs, etc.
2) Where in the world does have a proper legal status for co-operatives, and how does it benefit them?

To be clear, I like the idea, we just need to flesh it out a little. :hand: for now.

Floppy:
@digitalWestie are you up for expanding this a little? If not, it will be closed having passed the 90-day deadline, and is now languishing...

To reiterate, I like the idea, but I'm not sure what the legislation that's being proposed actually is, or is intending to fix.

digitalWestie:
Hey, if you're wondering what happened to this I decided it's probably better split out into multiple proposals. Will add when I get the chance.

Floppy:
thanks!

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/248), and results are set in the merge status checks below.

## Changes

PR #257, opened 2014-12-30
Enforce living wage

philipjohn:
An idea to enforce the minimum living wage.

### DIFF
--- a/manifesto/economy.md
+++ b/manifesto/economy.md
@@ -13,6 +13,8 @@ What policies should we propose to maintain a competitive economy that provides 
 
 The minimum wage should be set to the same value as the living wage (currently £14,458.50 based on 252 working days of 7.5 hours per day at £7.65), and automatically adjusted in line with that figure.
 
+Help will be given to companies to transition their staff to the minimum wage where their current business plan cannot immediately support a living wage. Companies that continue to make profits whilst paying below the living wage will be liable to fines of up to 100% of those profits, and the directors will be liable to personal prosecution with a penalty equivalent to their entire salary and any dividends, minus the living wage.
+
 ## Taxation
 
 ### Income Tax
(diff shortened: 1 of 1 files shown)


Floppy:
I'm still :+1: on this btw, although it may be superfluous. No harm in including it.

PaulJRobinson:
I like the idea of harsh penalties for failing to pay LW. But this proposal is a little over the top though isn't it? 100% of profits? 100% of salary? It's not a proportional punishment. Could we amend to something that doesn't sound quite so extreme. It just doesn't sound like we've even thought about it an appropriate punishment.

 Also there's no ability to scale the punishment up depending on the gravity of the offence, or any aggravating circumstances (such as committing a previous similar offence). If one worker in an organisation of 10,000 people is denied LW then the penalty is huge, but it's the same punishment if they deny LW to 5,000 of their workers, or if they've been fined three years in a row for doing the same thing. :-1:

philipjohn:
> It's not a proportional punishment.

Maybe. But here's the reality: income inequality is getting worse. More and more people are slipping below the poverty line. More and more people are having to get what work they can which often means low paid, insecure contracts. Meanwhile, profits at some of the biggest firms are rising, shareholders pockets are being lined ever more thickly.

This proposal provides in law a moral basis for shareholder accountability over workers pay. As a shareholder you should not be profiting from the underpayment of workers. It doesn't matter whether that's one worker or thousands - the crime is the same. Note that the exact wording is "up to" 100% - not exactly 100% in every case.

I'd be willing to add something here to the effect that the fine would be used to a) back-pay workers who haven't received the LW and b) used in the short-term grants scheme to help employees move up to the living wage.

Floppy:
Happy with "up to", yes. And adding those uses of the fine to tie into other policies is a nice idea.

philipjohn:
I've updated to add a bit about using the fines, and also to tweak the wording to say minimum wage given that IS the minimum wage.

Floppy:
:+1: 

Your thoughts, @PaulJRobinson?

digitalWestie:
:+1: yes to this

philipjohn:
@PaulJRobinson What do you think about the change above chap?

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/257), and results are set in the merge status checks below.

## Changes

PR #260, opened 2015-01-02
student president should be non-voting member on governing body

yellowgopher:
School Governance.  Advise student president should be a co-opted (non-voting) member on governing body to avoid any legal issues surrounding liability etc.

### DIFF
--- a/manifesto/education.md
+++ b/manifesto/education.md
@@ -39,7 +39,7 @@ Scottish authorities will no longer be legally obligated to appoint unelected re
 
 ## School Governance
 
-All state secondary schools will have an elected students council, and student president. The student president, who will need to be at key stage 4, will represent students on the board of governors.
+All state secondary schools will have an elected students council, and student president. The student president, who will need to be at key stage 4, will be a co-opted member on the board of governors.
 
 [^1]: [British Social Attitudes Survey](http://ir2.flife.de/data/natcen-social-research/igb_html/pdf/chapters/BSA28_12Religion.pdf) (pdf)
 [^2]: [Maintained Faith Schools, Department for Education](https://www.gov.uk/government/publications/maintained-faith-schools/maintained-faith-schools)


philipjohn:
Could you expand on the ramifications for the student president being a vote-wielding member? I'm genuinely unsure and as the original author of this bit of policy I'm keen to see it stay in with the intended effect (that of introducing students to genuine, effective democracy).

I'm worried that a non-voting role just wouldn't have the right clout, but it sounds like there might be big issues with that?

PaulJRobinson:
Happy with this. But a co-opted member isn't necessarily a non-voting member. If that's your intention with this proposal I think it should be explicitly stated in the text.

yellowgopher:
Generally co-opted members do not have formal voting rights but, you are correct, it would be up to the articles of the body in question whether this would be the case.  It can be stipulated if you think this needs to be clarified.

philipjohn:
If the intention *is* to make the student president a non-voting member I think that would take away the reason for this proposal in the first place - of introducing students to a tangible democratic process that gives them a voice.

Without having a vote, there's no point in the student president at all, nor the democratic system around it.

yellowgopher:
My concern would be potential issues surrounding liability of the student president being a full member of the governing body and able to have a final say in what happens.  Also there are issues that are dealt with by a governing body that may or may not be age appropriate for the student president - disciplinary hearings, abuse, headteachers' performance and pay etc etc.  If we decide to remove these areas from the student president's remit we then have "classes" of governor which goes against a key definition that each full governor is equal.  So, although I appreciate the reasons behind the proposal for the student president to be a full vote wielding governor, I think it is far better (and safer) to have them as a co-opted member that should be consulted on areas that are age appropriate (we may have to specify they must be consulted on a specific list of areas).  They will be introduced to the democratic process, governing bodies may be willing to give them a vote on limited areas (but this shouldn't be the law); it covers the main part of what is wanted but continues to protect the student president who would be a minor.

yellowgopher:
Ok, I think I have somehow managed to re-request this!  I have added a section about requiring the GB to consult a student president on certain matters - does this help?  Let me know if I can close this off...!?

philipjohn:
> able to have a final say in what happens
Why is that a bad thing?

> Also there are issues that are dealt with by a governing body that may or may not be age appropriate for the student president - disciplinary hearings, abuse, headteachers' performance and pay etc etc.
All of which is implementation-specific, and would be dealt with by the relevant department when consulting before implementing such a system, as such, isn't something we should get bogged down in when creating this vision.

> should be consulted
...and inevitably ignored, introducing the students to a democratic system that ignores them, setting them up for a life of political apathy, not engagement.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/260), and results are set in the merge status checks below.

## Changes

PR #262, opened 2015-01-02
Include all public institutions in the FOIA

yellowgopher:
Change Royal Secrecy to Freedom of Information - to include any exempt institutions in the FOIA so they have to respond to FOI requests (bearing in mind that there are criteria whereby FOI requests can be refused anyway!)

### DIFF
--- a/manifesto/democracy.md
+++ b/manifesto/democracy.md
@@ -101,31 +101,26 @@ Ban unelected Lords from serving in Government.
 
 Cabinet appointments will be subject to a vote of approval by the House of Commons.
 
-## Royal Secrecy
+## Freedom of Information
 
-Fully include of the monarchy within the Freedom of Information Act (FOIA), so that the royal household is legally obliged to respond to information requests[^2].
+Fully include any exempt institutions within the Freedom of Information Act (FOIA), so that they are legally obliged to respond to information requests.
 
-Repeal existing exemptions from the FOIA that allow communications between other government bodies and the members of the royal family to be kept secret.
+## The Constitutional Monarchy
 
-Adopt full disclosure of royal lobbying and influence, including disclosure of meetings between members of the royal family and ministers.
+Although the Constitutional Monarch inherits the role of Head of State by just being born, there is a great deal of benefit in maintaining it in this role.  The Constitutional Monarchy apolitical, is by nature distant from the day to day running of the country, provides continuity and history and exisits for a very limited set of reasons (to carry out important roles such as the appointment of Prime Ministers, approving legislation and bestowing honours).  It is a strong tourist attraction for the country and carries out a large amount of charitable work.  Being a Constitutional Monarchy it is controlled by the constitution (and in the case of United Kingdom, convention) and owes it's continued existence to parliament and, ultimately, the people.  
 
-## Republic
+Because a Constitutional Monarch has a limited set of reasons to exist, it allows the incumbant to fully learn and understand their role in a way that an elected president wouldn't have the time to do so.  The Consitutional Monarch lives the role.
 
-To become a truly democratic country, the UK should become a republic. As a long-term goal, the monarchy would be replaced with an elected President who will act as Head of State. We recommend the [model proposed by Republic](http://republic.org.uk/What%20we%20want/In%20depth/The%20Case%20for%20a%20Republic/Policy/index.php).
+The partial seperation of the Constitutional Monarch as Head of State from day to day politics is a very secure way to ensure that the country remains stable.
 
-The President would be mainly a ceremonial position, and should be apolitical and not affiliated to any party. The Prime Minster would remain as Head of Government, and would be appointed by the President after a General Election, as is the case with the Queen now. The President will not be involved in the legislative process.
-
-Presidential terms of office will be fixed at five years, with a maximum of two terms to be served by an individual.
-
-The President will be equal before the law, and will not be protected by [Sovereign Immunity](https://en.wikipedia.org/wiki/Sovereign_immunity#United_Kingdom). The President will not be constitutionally linked to any faith.
+Having a Constitutional Monarchy does require a small amount of democratic sacrifice but, as a country, we already have the checks and balances in place to ensure the Monarch doesn't exceed certain boundaries - and we have implemented these on many occassions over the centuries.  For this reason there is no need to change the role of the Constitutional Monarch as Head of State.
 
 ## Democracy Research
 
 Investigate the feasibility of direct digital democracy - using online tools to gradually replace representative democracy with increased direct participation in the legislative process by all electors.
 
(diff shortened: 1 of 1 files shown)


Floppy:
Thanks! There are two distinct things here, one of which will easily be accepted, and one which is a much more major change. Any chance you could separate the two?

yellowgopher:
Yeah, probably should have done that originally tbh. Do you know if there is any easy way to undo changes or do I need to resubmit?

tmtmtmtm:
The first should probably also spell out a little more what "any exempt institutions" is to include. Stated as as present it's rather vague and almost certainly over-broad.

Floppy:
Probably easiest to resubmit, if that's OK? Otherwise I can try to split it, but I'm not sure when I'll get time...

philipjohn:
If you submit one of them separately I'll modify this one tonight to remove it. That should sort it.

yellowgopher:
OK, remove the Constitutional Monarchy changes from this  - this will just be the change in FOI.

Not sure how vague this is - my recommendation would be that there should be no exempt organisations or institutions.  The idea that all FOI requests should be considered on their merits...?

tmtmtmtm:
> my recommendation would be that there should be no exempt organisations or institutions.  

I'm confused. What do you mean by organisation, here?

yellowgopher:
Any institution or organisation.  The original manifesto clause pointed out that the Monarchy shouldn’t be exempt.  I suggest expanding that to include anyone or anything that could have an exemption in law – Monarchy, MoD etc.

philipjohn:
So, basically, remove from the act any mention of exemptions so that all public bodies are subject by default, right?

tmtmtmtm:
The vast majority of organisations in the country are exempt, by virtual of not being government bodies. I'm assuming you're not trying to extend FOI here to the private sector, for example?

philipjohn:
@tmtmtmtm As the act only applies the public authorities, removing exemptions would still only affect public authorities

tmtmtmtm:
> So, basically, remove from the act any mention of exemptions so that all public bodies are subject by default, right?

The FOI Act doesn't really work like that. It's based around a positive list, rather than a negative one:
  http://www.legislation.gov.uk/ukpga/2000/36/schedule/1

philipjohn:
That first paragraph does say all public authorities though, and beyond that are specific inclusions and relevant exclusions.

tmtmtmtm:
> As the act only applies the public authorities, removing exemptions would still only affect public authorities

But you still have to explain what a 'public authority' is. If you're not going to do it with a big long list (as at present), you need to give a workable definition that makes it obvious whether any given body is subject or not.

philipjohn:
Yep, which is handled in FOIA by the section you've linked to. That's implementation detail though - the stuff for lawyers drafting legislation to tie themselves in knots over - rather than the principle of the policy proposed here.

tmtmtmtm:
I think we're talking past each other here. The Royal Family, for example, is currently exempt from FOI simply because it doesn't fall under any current definition of being a public authority — not because there's an explicit exemption in the FOI Act removing it. [1]

I'm not looking for precise legislative wording here — I'm simply looking for a high-level statement that says "An organisation is a Public Authority for the purpose of FOI if ... [what?]".

Currently the answer to that is "if it appears in this big list". We either need to maintain that approach and simply extend that list, or else offer a replacement definition.

--

[1] There *has* been a recent change to add an exemption for information relating to the Royal Family — but that's about, for example, asking the Cabinet Office about letters that Prince Charles has sent them, not about asking Prince Charles himself for them. I'm all in favour of undoing that exemption, but that's about categories of information that should be released, not about changing the list of public bodies.

philipjohn:
Ah! I see your point :) I don't have a good answer to that... :/

philipjohn:
> OK, remove the Constitutional Monarchy changes from this

@yellowgopher Turns out I can't edit this, which actually makes sense. You can though, by clicking the edit button here: https://github.com/yellowgopher/manifesto/blob/patch-5/manifesto/democracy.md

Floppy:
I'm definitely in favour of the core FOI idea here, we just need to sort out the details, I think. Yay!

tmtmtmtm:
A good way of working out details would be to come up with some bodies that aren't currently subject to FOI that we think should be, and then working out whether there's a pattern that can be captured, or whether they simply need to be itemised for inclusion.

Floppy:
@tmtmtmtm for now, would you accept a statement of general principle that all public bodies, including the royal household, should be subject to FOI? We can work out the detail later on if necessary.

tmtmtmtm:
> would you accept a statement of general principle that all public bodies, including the royal household, should be subject to FOI? 

No, because it's begging the question. 

The law already says that all public authorities are subject to FOI.

So us saying likewise seems to be either simply restating what the law currently is, for no obvious purpose, or hinting at something we want to change without stating what that is. 

If all we want to change is that the Royal Household should be subject to FOI, then let's simply say so.

But if we want to change something wider than that, then we need to explain it.

philipjohn:
Is it your view then that the current wording is fine and this proposal isn't needed?

tmtmtmtm:
My issue with the proposals here is simply that they're Not Even Wrong, as they seem to be predicated on the idea that (a) there's a clear concept of what a public authority is, and (b) some of those public authorities are currently exempted from FOI law. But this isn't how the FOI act works, so none of the proposed fixes make much sense. 

If someone can point to, or come up with, a plausible definition of a public body, that would let us neatly claim "FOI should apply to [everyone who meets these criteria]", then, sure, I'd probably support that. But in the absence of that, I think we should simply list anyone we think should be added to Schedule 1.

Currently the only addition mentioned here has been the Royal Household (which I would support). Personally I'd like to also see many of those listed at https://www.whatdotheyknow.com/body/list/foi_no [1] become subject to FOI, but I'm not sure how many of them I feel strongly enough about to propose them individually at a manifesto level.

----

[1] This is the tag we (the WDTK administrators) give to organisations there that are not currently subject to FOI, but which either agree to act as if they were, or which we think *should* be subject to it. Our criteria for that are listed at http://foiwiki.com/foiwiki/index.php/WhatDoTheyKnow_Policy:Authorities_not_subject_to_FOI

Floppy:
Oh, now that's a *really* useful list. How about if we said something like:

```
Include all bodies with public responsibilities under FOI, including but not limited to [a few examples]. For more examples, see http://foiwiki.com/foiwiki/index.php/Organisations_and_officials_with_public_responsibilities_that_are_not_subject_to_the_Freedom_of_Information_Act
```

I'd include as examples:
* The Queen
* The Royal Household
* The Press Complaints Commission
* The Advertising Standards Authority
* Utility companies (water, sewage, energy, transport)

yellowgopher:
I would be happy with that.  It sets a principle towards which we can aim.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/262), and results are set in the merge status checks below.

## Changes

PR #263, opened 2015-01-02
Retain constitutional monarchy

yellowgopher:
Change Republic to Constitutional Monarchy

### DIFF
--- a/manifesto/democracy.md
+++ b/manifesto/democracy.md
@@ -109,15 +109,15 @@ Repeal existing exemptions from the FOIA that allow communications between other
 
 Adopt full disclosure of royal lobbying and influence, including disclosure of meetings between members of the royal family and ministers.
 
-## Republic
+## Constitutional Monarchy
 
-To become a truly democratic country, the UK should become a republic. As a long-term goal, the monarchy would be replaced with an elected President who will act as Head of State. We recommend the [model proposed by Republic](http://republic.org.uk/What%20we%20want/In%20depth/The%20Case%20for%20a%20Republic/Policy/index.php).
+Although the Constitutional Monarch inherits the role of Head of State by just being born, there is a great deal of benefit in maintaining it in this role.  The Constitutional Monarchy apolitical, is by nature distant from the day to day running of the country, provides continuity and history and exisits for a very limited set of reasons (to carry out important roles such as the appointment of Prime Ministers, approving legislation and bestowing honours).  It is a strong tourist attraction for the country and carries out a large amount of charitable work.  Being a Constitutional Monarchy it is controlled by the constitution (and in the case of United Kingdom, convention) and owes it's continued existence to parliament and, ultimately, the people.  
 
-The President would be mainly a ceremonial position, and should be apolitical and not affiliated to any party. The Prime Minster would remain as Head of Government, and would be appointed by the President after a General Election, as is the case with the Queen now. The President will not be involved in the legislative process.
+Because a Constitutional Monarch has a limited set of reasons to exist, it allows the incumbant to fully learn and understand their role in a way that an elected president wouldn't have the time to do so.  The Consitutional Monarch lives the role.
 
-Presidential terms of office will be fixed at five years, with a maximum of two terms to be served by an individual.
+The partial seperation of the Constitutional Monarch as Head of State from day to day politics is a very secure way to ensure that the country remains stable.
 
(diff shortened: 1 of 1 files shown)


philipjohn:
> already have the checks and balances in place to ensure the Monarch doesn't exceed certain boundaries

[Clearly not](http://www.theguardian.com/uk/2013/jan/14/secret-papers-royals-veto-bills).

Big :-1: from me on this, as I think it directly conflicts with our first value (my emphasis):

> Open, transparent and *accountable*, with no hiding place for backroom influence

An unelected, hereditary head of state is clearly unaccountable and we're seeing more and more evidence of political interference from the current and future head of state.

Floppy:
Likewise, this is a :-1: from me, as I believe that a truly democratic society can't have a head of state that's not accountable to the public. It belongs in the 19th century, not the 21st. If we're looking forward, we need to be moving beyond such relics of the past.

There is a nice monarchy mythbuster on the Republic site that responds to many of the points raised in this PR, which I quite like: http://republic.org.uk/what-we-want/monarchy-myth-buster

PaulJRobinson:
I've come full circle on this since we started this project (the real value of political discourse!) and no longer support a constitutional Monarchy. Much rather have an elected (but ceremonial) monarch on single fixed term who operates like a national village Mayor opening fetes and museums etc. And their extended family won't get to tag along for the taxpayer funded ride either. :-1:

yellowgopher:
The value of political discourse indeed.  I for one am not convinced of the republican route and I don't really believe it is a battle that needs to be fought.  I also think there is a lot of misinformation banded about by lobby groups such as Republic that do not state all the facts - such as the cost of Monarchy in the UK and the "extended family" tagging along comment above!  If we have to go down such a route I would much rather we do the research first rather than relying on others - I think I will start reading up about all of this!

philipjohn:
> such as the cost of Monarchy in the UK

Do you dispute the research that Republic have done about the cost of the Monarcy, and if so, do you have evidence to the contrary?

yellowgopher:
Well, yes, because it is too simplistic!  It makes big assumptions as to what it applies to the cost of having a monarch.  For example, the Crown Estates are not  "owned" by the Monarch, nor are they "Public owned".  They are owned by the Crown and managed by the Crown Estate Commissioners - which is an independent body.  Currently the agreement is that Crown Estate revenues are passed to the Treasury in return for a Sovereign Grant (c.15% of revenues by the way - so the Treasury makes money here).  The problem here is that IF there is no Sovereign Grant then that side of the agreement is broken and one would presume we would revert back to pre-Civil List times (when the Monarch had control of the Crown Estate).  It would certainly be a challenging legal battle!  The residual body that replaced the Crown in a republic would arguably have more right to the Crown Estate portfolio than the government or the public as a whole - so we cannot guarantee any money would be saved/made from this.  But there are other arguments too.  Councils pay for royal visits - they would pay for presidential visits too.  The Met cover the cost of security for the Royal Family - I presume they would continue to provide security for any member of society afterwards (including a new presidential family and associated staff).  So I don't see how a cost benefit would occur...!

yellowgopher:
I have more sympathy for arguments that suggest a President could do just as good a job as a Monarch.  I agree, they could.  However you are not going to get away from an elitist culture by abolishing the Monachy - by definition those in a position to apply for the job will either have the support of the establishment, money, power, influence etc.  And abolishing the Monarchy in itself doesn't necessarily reduce the amount of power politicians have - that they have a lot stems as much from our unwritten constitution and the use of constitutional conventions as it does from the royal prerogative, reserve powers etc (some or all of which may remain under a Present!) - I know that this is something else that is currently in the manifesto...!

But!  I guess this is all beside the point really.  My greatest issue of all is that a political force that is positioning itself as a conduit of the people (i.e. the people form the manifesto) is making such a one sided statement at the beginning of the formation of its manifesto!  Surely the position on the Monarchy should be neutral and allowed to evolve one way or the other?  I appreciate the members of the party lean towards republicanism but that's not necessarily the "view of the people" is it?

yellowgopher:
Well I guess this is blocked now and dead in the water anyway...

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/263), and results are set in the merge status checks below.

## Changes

PR #267, opened 2015-01-10
Scrap Vehicle Excise Duty

Floppy:
Put the tax on fuel instead, saving money on admin, and still having the desired effect of taxing emissions.

### DIFF
--- a/manifesto/transport.md
+++ b/manifesto/transport.md
@@ -13,6 +13,14 @@ What policies could improve public and private transport infrastructure?
 
 No new fossil-fuel powered vehicles may be sold in the UK after after 2029.
 
+## Scrap Vehicle Excise Duty
+
+As Vehicle Excise Duty is associated with vehicle emissions, it seems fairer to tax
+the fuel directly. VED should be scrapped, and fuel duty increased to cover the reduction
+in income. Overall, the tax levied should be lower, due to the removal of the need to administer
+VED, though the tax will fall more heavily on those who use more fuel, and thus create more
+carbon emissions.
+
 ## National Infrastructure Organisations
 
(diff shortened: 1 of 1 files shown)


philipjohn:
:+1:

otfrom:
but doesn't put a tax on the emissions of creating the vehicle

Floppy:
I don't think VED does that now, and I'd prefer that embodied carbon cost (for *everything*) would be included in a proper carbon pricing regime. After all, under VED, a huge electric vehicle would still be zero-rated, as it's based on exhaust emissions.

tmtmtmtm:
I'd prefer the argument for this was backed up by something stronger than "it seems fairer". This is presumably something that there's already been quite a lot of research into, and that different countries will have deliberately chosen different approaches to. Is there a good summary of the issue somewhere?

Floppy:
Agree the "seems fairer" wording is vague. Is there a general principle of taxing the thing you're measuring directly as opposed to indirectly? What's that called? Direct/indirect is something else...

Floppy:
A summary of "road taxes" from around the world: https://en.wikipedia.org/wiki/Road_tax. Most seem to be based on emissions, but such a tax is nowhere near universal, and some, such as France's, have been abolished and replaced with more direct measures.

Floppy:
I suspect there is a different PR to be made on tax philosophy in general, but I'd prefer not to get into that here.

PaulJRobinson:
:+1:

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/267), and results are set in the merge status checks below.

## Changes

Issue #71, opened 2014-01-13
Ask for evidence


philipjohn:
Quick suggestion: we should collectively take a cue from Sense About Science in formulating the manifesto and [Ask For Evidence](http://www.senseaboutscience.org/pages/a4e.html).

tmtmtmtm:
I definitely like the idea of another vote type. There are quite a few proposals for which I don't particularly want to vote No, as I'm not necessarily against the idea, but where I believe there needs to be much more explanation given. The changes in #270 mean I'll probably *need* to start voting No on them, but I'd prefer to have an alternative vote type instead.

Floppy:
What do you think the result of an evidence vote should be? Is it effectively the same as the block, but without the negative connotations?

Floppy:
I will attempt to choose an icon from http://www.emoji-cheat-sheet.com/. Fun.

Floppy:
Probably :question: is most appropriate :)

tmtmtmtm:
> Is it effectively the same as the block, but without the negative connotations?

Kind of. And maybe it's just a matter of adjusting my own understanding of the meanings, but I see a 'no' vote as "I think this is a bad idea", whereas this would be "I don't know yet whether this is a good idea or a bad idea, and there isn't enough here to decide".

tmtmtmtm:
If we wanted there to be *some* distinction from a 'No', then perhaps it isn't a complete block — with sufficient (FSVO) Yes votes, the PR could still be approved anyway (though perhaps only manually, rather than automatically?)

I suspect we want to see how it plays out in a few cases first, though.

philipjohn:
> Any contributor can block a merge by including a ‘thumbs-down’ symbol, as long as they include constructive reasoning and alternative proposals.
http://openpolitics.org.uk/contributing.html

I think it'd be worth being clearer on the rules in this respect though (in a separate PR!)

> If we wanted there to be some distinction from a 'No', then perhaps it isn't a complete block

The hand emoji would perhaps work well for that? A hand could perhaps cancel out a thumbs up?

tmtmtmtm:
> A hand could perhaps cancel out a thumbs up?

There's something nice about that, though I'm worried about the case where a few people vote yes straightaway on gut instinct. Then (especially in a post #270 auto-committing world), simply saying "Hey, there's no evidence provided here at all" with a hand, or question mark, or whatever, wouldn't be enough.

[...]

PR #270, opened 2015-01-13
Update contribution rules

Floppy:
This changes the contribution rules in a couple of important ways, and improves the layout and wording of the rest of the guide. This is based on experience of the system over the last year or so, and a desire to make things move more quickly.

Changes to voting:

1. Only two existing contributors are needed to pass something. This is no longer based on a percentage of existing contributors. We have enough active members that incompatible ideas will be blocked before being merged.
2. Reduce minimum open vote time to a week. Two weeks is long enough for things to grind to a halt, one week should increase momentum. In my experience, consensus forms quickly or not at all. A week is enough.
3. Changes will be closed after 3 months if they're not merged, to keep the backlog clean and remove the interference from other old PRs hanging around.
4. Changes will be automatically merged on passing by the script so that admins can't sit on approved changes if they're not 100% happy with them. Block or pass.

We should also be encouraging blockers to counter with their own proposal, or if blocking for reasons of "this isn't detailed enough", to allow the change and build on it in future PRs.

This PR will be merged using the existing system, which means we need 5 yes votes to pass it.


Floppy:
Pinging @PaulJRobinson @philipjohn @digitalWestie @tmtmtmtm @yellowgopher @rufflemuffin @BenTrimble @frabcus

As our most recent contributors, you guys are most likely to have opinions on this change, and we need your votes :)

tmtmtmtm:
> "We should also be encouraging blockers to counter with their own proposal, or if blocking for reasons of "this isn't detailed enough", to allow the change and build on it in future PRs

I disagree with this: I thought the aim was to avoid policy based purely on ideology? Where no justification whatsoever is provided for a policy, for example, it should *always* be acceptable to say "this needs more information". 

(The version in the PR itself is somewhat better as it refers to "partial *improvements*". (Emphasis mine))

Floppy:
Yes, take the version in the PR as the important one, but I see what you mean there. I really want to say that if we know something is basically good, but would benefit from expansion, we encourage adoption and improvement in subsequent PRs. Iteration over big bang. Does the proposed text do that, ignoring my bastardisation of it in the description, do you think?

tmtmtmtm:
I think the text in the PR itself is OK; but it would be better if applied in parallel with something for #71

tmtmtmtm:
> If the script detects that a change has been passed, it will be merged automatically

I suspect there should be at least some delay between the final 'yes' vote, and the auto-merge.

philipjohn:
Firstly, +10 manifesto points for mentioning #71 (there's no badge/prize/whatever, sorry)

I thought briefly about whether we could do something around that. We would perhaps have to rely on the voters to be blocking something that isn't evidenced well (within reason) though.

> I suspect there should be at least some delay between the final 'yes' vote, and the auto-merge.

My interpretation is that a change would get merged after one week if it's passing the criteria. Are you suggesting that if, for example, the required votes happen on day 7, the merged should be delayed by X amount of time. I.e. "merge after 7 days if last vote was less than X ago"

tmtmtmtm:
yes — or even later. I'm thinking here of a scenario where, after 7 days, there has only been one yes vote. Then, a week later (say) there's a second yes vote — at which point (by my reading of this), the change will get merged straightaway. I'd like to see even a 24-hour delay or so here, to allow people time to react to it. Yes, everyone should have been paying constant attention, and all that, but still...

philipjohn:
> Yes, everyone should have been paying constant attention

Ha, it's all too easy to glance at a PR and then forget about it though, so it does make sense.

24 hours sounds good. I wonder if we could rig up the option for folks to be notified the moment a PR is 24hrs away from being merged.

Floppy:
OK, yeah, fair enough. We could have the auto-merge have a secondary delay, though of course a manual merge could happen quicker. Do we want to have a minimum delay before merge, or would it be OK to do immediately on passing if a human does it?

Floppy:
OK, I've updated the text to include a 24 hour delay after passing before auto-merge, and to make abstentions cancel an agreement. I think we will need to pass this under the proposed rules rather than the existing ones though, otherwise we'll stall completely.

philipjohn:
:+1:

Floppy:
@tmtmtmtm do you think this is OK now?

tmtmtmtm:
Hmm. I suspect that someone looking at it fresh will likely be very confused as to why the 'hand' is a -1 count, and the '-1' is a halt...

Floppy:
That's true, but unless we integrate to a more complex voting system, that's probably the best we can do. Unless we change block to something like :x:, perhaps.

tmtmtmtm:
By 'more complex', you mean for the voter? Or for the code of the votebot? It's certainly not ideal to reverse the meanings, but if we *were* to do so, then now is likely to be better than at any time in the future. Making '-1' actually *mean* -1, and then taking 'hand' or 'X' to mean stop seems like it could avoid significant confusion later, and (with suitable handwaving) could be handled with date conditionals in the code for a while.

Floppy:
I meant something like Loomio. I don't mind complex code in the votebot. I'd rather have a new set of symbols to avoid confusion though, especially when reading back. Do you think we could proceed with the current symbols to get the rules in place, and change the symbols as a separate PR?

tmtmtmtm:
Ah, gotcha. That's a good point about being able to read historic PRs, so a slightly hesitant :+1: for now from me[1]. It would be good if we could produce a list of changes that will be affected by this, before merging though, so that people have a flurry of activity on ones that people have forgotten about.



[1] Aside from the oddness of the back-to-front signals, I'm also a little concerned that to "make things move more quickly" is potentially a step in the wrong direction. But I'm not sure how much of that is simply to do with mismatched expectations as to what a process like this *should* (IMO) be, and my thoughts on this aren't quite coherent enough yet to really express them in any detail. But I can live with that for now :)

digitalWestie:
Seems legit :+1:

PaulJRobinson:
:+1:

philipjohn:
Pinging @yellowgopher @rufflemuffin @BenTrimble @frabcus as we just need another vote or two to get this in (if you agree)!

yellowgopher:
OK, sounds reasonable to me. Need to get my head around this. Can I vote? Thought I wasn't eligible yet...!

yellowgopher:
OK, sounds reasonable to me. Need to get my head around this. Can I vote? Thought I wasn't eligible yet...!

philipjohn:
@yellowgopher My mistake, looks like you don't have an accepted proposal yet :(

Floppy:
We have 4 votes out of 5 needed for this, but the fact we can't get another one to just shows how necessary it is. I'm going to be controversial and merge this with the 4 votes we have now, given that we've had a good debate and there is no opposition that we can get people to express. Will do so in a few hours, so you have chance to object to my blatant subversion of democracy.

Floppy:
@mikera I thought you'd left us forever :smile:, but if you're around, I'd love to get your vote on this proposal to change the voting rules. We're slowing down a bit here, and need to re-inject some agility and energy!

mikera:
:+1: I generally think getting changes in faster and iterating is a good idea.

Floppy:
BOOM. And we're done. Thanks @mikera.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/270), and results are set in the merge status checks below.

## Changes

PR #272, opened 2015-01-18
Local hustings for general elections

philipjohn:


Floppy:
:+1: definitely. There should be some official hustings hosted by the organisation running the vote.

PaulJRobinson:
:+1:

tmtmtmtm:
Is this a responsibility to ensure that at least one public husting takes place? I'm not sure the current wording says that (You can enable something without it actually happening.)

Floppy:
Perhaps it should be a "duty to ensure that at least one independent public hustings takes place and is promoted to all constituents". That way they have the option to work with other organisations if that's more effective.

Floppy:
I've followed this up with #292 which refines the wording slightly.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/272), and results are set in the merge status checks below.

## Changes

PR #279, opened 2015-02-15
Give the vote to everyone with an NI number

Floppy:
Allows votes at 16, and allows people working here who are paying tax but aren't citizens to vote anyway.

### DIFF
--- a/manifesto/democracy.md
+++ b/manifesto/democracy.md
@@ -23,7 +23,7 @@ Introduce voting by proportional representation using the Single Transferable Vo
 
 Investigate the feasibility of electronic voting at all elections, with a view to increasing accessibility and turnout.
 
-Add anyone over the age of 16 with a National Insurance Number to the electoral roll for voting in General, Scottish, and European Elections (i.e. to legislative bodies), as well as all national and local referenda. This will mean that anyone over 16 may vote, as well as anyone working in the UK even if they do not have full citizenship. In the words of the old American colonists, "no taxation without representation".
+Allow anyone over the age of 16 with a National Insurance Number to register to vote in General, Scottish, and European Elections (i.e. to legislative bodies), as well as all national and local referenda. This will mean that anyone over 16 may vote, as well as anyone working in the UK even if they do not have full citizenship. In the words of the old American colonists, "no taxation without representation".
 
 In addition to the above, allow voting at Local Elections and to the Welsh Assembly (i.e. to non-legislative bodies) at age 15, to encourage youth participation in democracy prior to enjoying full voting right at 16.
 
--- a/manifesto/democracy.md
+++ b/manifesto/democracy.md
@@ -23,9 +23,9 @@ Introduce voting by proportional representation using the Single Transferable Vo
 
 Investigate the feasibility of electronic voting at all elections, with a view to increasing accessibility and turnout.
 
-Reduce the age of voting at General, Scottish, and European Elections (ie to legislative bodies) and all national and local referenda to those aged 16 years and above.
+Add anyone with a National Insurance Number to the electoral roll for voting in General, Scottish, and European Elections (i.e. to legislative bodies), as well as all national and local referenda. This will mean that anyone over 16 may vote, as well as anyone working in the UK even if they do not have full citizenship. In the words of the old American colonists, "no taxation without representation".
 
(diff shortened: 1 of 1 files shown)


tmtmtmtm:
If, say, a twelve-year old starts their own online business and earns enough from it to pay tax, will they also gain the right to vote?

Floppy:
I don't think they will have an NI number at that point, so not under this proposal, no. But that is a whole other interesting question you pose there.

Floppy:
Maybe this should still mention the minimum 16yo requirement.

Floppy:
tweaked slightly, because apparently NI numbers can be issued a good while before age 16.

sanjaypoyzer:
You actually get your NI number a little while before you're 16 (I remember getting mine a good 6 months before my 16th) so that age requirement does make quite a difference if one of the big advantages of this is making voter registration easier.

From a quick search, I can't see anything about under 16 year old entrepreneurs getting their NI number early. Where did you see it @tmtmtmtm? I did see something once about under 16 year olds using their parent's child benefits number if they need to pay tax, but I can't find it now unfortunately...

tmtmtmtm:
I'm OK with the principle of extending who has the right to vote like this, but the wording makes it sound like this is also bringing in automatic voter registration. Is that part of the intention?

tmtmtmtm:
@sanjaypoyzer my comment wasn't really about the NI number part itself (AIUI, you don't pay National Insurance until you're 16, even if you're paying income tax). It was more a riff on the "No taxation without representation" aside.

sanjaypoyzer:
That's what I thought was being proposed. I think automatic voter registration would be great. I can't see any other outcomes other than saving public money and increasing voter turnout.

philipjohn:
> also bringing in automatic voter registration. Is that part of the intention?

Hope so! :+1: 

@tmtmtmtm @sanjaypoyzer Don't forget to vote on this!

tmtmtmtm:
I'm waiting for @Floppy to confirm whether or not this is meant to be smuggling automatic voter registration in as well...

Floppy:
Ah, sorry. I did see automatic voter registration as a positive side effect, yes, though I don't actually know if it would work because I have no idea if NI numbers are linked to home addresses. But I do think it would be a beneficial side effect.

tmtmtmtm:
I'd say that it's worth adding automatic voter registration as a standalone proposal. I think they're separate-enough concepts, and I'm a little wary of having them so tightly combined.

Floppy:
OK, that's reasonable. Let me edit.

Floppy:
Try that on for size. I already know the next PR I'll put in to change it to automatic, but as you say - separate issue.

tmtmtmtm:
Looks good.

However what's with explicitly listing Scottish elections here? This seems to be enshrining the position that this should be a Reserved Matter, though it's unclear whether that's deliberate or not. It's also a little odd to explicitly list Scotland and Wales, and not Northern Ireland.

I know that this is mostly just because it's adapting text that was already there, but it seems like it should still be cleared up. I'm not sure I'd be happy with imposing this change into the Scottish Parliament or Northern Irish Assembly if it were against their wishes.

Floppy:
The Scotland / Wales split wording was in the text previously, though I'm not sure it's right, as the Welsh and NI assemblies do have legislative power. That might need fixing in a couple of places by someone who knows the precise answer, but can it be done separately?

tmtmtmtm:
I'm not sure if it can be a completely separate change, as this proposal isn't just sticking with what came before, but going further.

In Northern Ireland, for example, the Assembly is explicitly in favour of reducing the voting age, but I'm not so sure whether it would be in favour of *this* change too, as there are presumably all manner of implications in relation to the large number of cross-border workers, who live in the Republic, but work in the North.

PaulJRobinson:
I believe peers have NI numbers, but they don't get a vote (as they aren't 'Commoners'). Similarly I assume Royalty have NI numbers as lots have served in the Armed Forces and will receive a wage from them. Should they get the vote? I don't think they should. I think there has to be a better arbitrator of who gets the vote than NI numbers.

Floppy:
I don’t see any reason to withhold votes from peers or royalty; they are citizens after all, and given the rest of our plans, they will be just the same as everyone else at some point. What’s your rationale for withholding the vote from them?

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/279), and results are set in the merge status checks below.

## Changes

PR #283, opened 2015-02-18
Removal of party politics from Town Councils

andrewdwilliams:
Most of the entire point of Flatpack Democracy is that local government is outdated and we need movements such as Independents for Frome to inject some life into it. Most of the cause of this is the "irrelevant and corrosive diversion" that is party politics at local level. Creates unnecessary groupings and tribal conflicts rather than helps to get the best for the place the Town Council represents.

### DIFF
--- a/manifesto/local_government.md
+++ b/manifesto/local_government.md
@@ -14,4 +14,8 @@ Publish full minutes (or transcripts if available), audio recordings, and voting
 
 ## Waste and Environment
 
-The Localism Act 2011 repealed powers under The Climate Change Act 2008 which gave councils the ability to charge fine or introduce tariffs regarding household waste. Local authorities are primarily responsible for waste collection and recycling. Consequnetly, local government should have the power to issue financial discencintives (and/or incentives) to encourage recycling and effective waste collection.
+The Localism Act 2011 repealed powers under The Climate Change Act 2008 which gave councils the ability to charge fine or introduce tariffs regarding household waste. Local authorities are primarily responsible for waste collection and recycling. Consequently, local government should have the power to issue financial disincentives (and/or incentives) to encourage recycling and effective waste collection.
+
+## Removing Party Politics from Town Council
+
+Town Councils, which are effectively Parish Councils in all but name, have one distinct difference in that Councillors on Town Councils stand for political parties, whereas all Parish Councillors are not aligned. As Peter MacFayden notes in his book [Flatpack Democracy](http://www.flatpackdemocracy.co.uk/ "Flatpack Democracy"), "at local levels of representation, party politics as practiced by the current political parties are an irrelevant and corrosive diversion." We advocate the removal of party politics from Town Councils in the same way they are removed from Parish Councils.


andrewdwilliams:
I also made a few spelling corrections in the above statement on Waste and Energy or whatever.

philipjohn:
> Councillors on Town Councils stand for political parties, whereas all Parish Councillors are not aligned

This is new to me - is that defined in legislation somewhere? I love your proposal, but want to make sure we have the evidence to back up the need.

On a side note, I love your principles for Independents for Frome. Something New has a similar policy on candidates - while a party, we [ask that candidates](http://www.somethingnew.org.uk/standing_for_election) act in the interests of their electors, just within the wider values of the party.

Floppy:
It's not in legislation that I know of, but it is a convention certainly. Certainly town councils like @PaulJRobinson's are party-aligned, even though they are basically the same level of government as parish councils, so I think it's just a convention.

Floppy:
:+1: btw

tmtmtmtm:
Can someone explain what this would actually mean in practice?

andrewdwilliams:
From the searching I've done, I can't find it in legislation, but the fact of the matter is, Parish Councils and Town Councils, which are the same in all but name, have one other distinct difference: one is occupied by non-aligned Councillors, the other is occupied by party-aligned Councillors. I'm not sure why. But I imagine it will be in legislation somewhere.

What exactly I'm trying to have done is to remove party politics from Town Councils so that it's impossible to stand for a political party in a Town Council election. It's as simple as stopping party politics from playing a part in Town Council affairs, as it can corrupt and spoil what could otherwise be rational local decision making with tribal groupings. It wouldn't be hard.

I'm not advocating removing Town Councils as a whole, as Town Mayors can be an important civil function and attraction. But removing party politics from Town Councils seems like a sensible thing to do.

tmtmtmtm:
I also still don't understand what is being suggested, or why. What does "removing party politics from Town Councils" actually entail? Are you saying that anyone who is a member or supporter of a political party is disqualified from running for office? Saying that they can be so, but only if they keep it secret? Even if you remove parties from the electoral process, what do you do post-election? If half the councillors turn out to be members of a well-organised party, how do you prevent them from operating as a bloc? The idea that "stopping party politics from playing a part in Town Council affairs ... wouldn't be hard" seems rather far-fetched to me.

And if it's possible (and good) for Town Councils, why limit it to there? Why not City Councils or County Councils?

andrewdwilliams:
In the same way that party politics is removed from Parish Councils. At such a basic, local level, having a party allegiance is unnecessary - it's pointless - and it causes tribal groupings where people vote on issues because of their party allegiance, not because of the merits of said issues. 

On Parish Councils, members of political parties are not disqualified from office. I believe that Parish Councillors that are members of political parties do declare this in their Register of Interests, however. Naturally, groupings on Councils may arise, but if traditional party politics was kept out the equation, these groups wouldn't be voting on issues "Because I'm a Tory," it would instead, probably, be because of genuine concern.

You could of course bring the same point to City Councils, District and Borough Councils, County Councils and even the House of Commons, if you were so keen on the issue. But after Town Council level, you get to the point where the number of Councillors on the Councils brings in an almost necessity of groupings to be able to form the cabinet and govern the Council properly.

I found this extract from an essay, ["Contemporary issues in the government and administration of Latin American mega-cities,"] (http://archive.unu.edu/unupress/unupbooks/uu23me/uu23me09.htm) "most voters choose their councillors according to the political party each represents; they know or care little about the person they are electing. Thus, the legitimacy of government varies, as does the form of rationality which will govern an individual's behaviour once in office."

We have given party politics a century to rule our Town Councils, and I think most people would agree that they have done a dismal job of it.

tmtmtmtm:
> In the same way that party politics is removed from Parish Councils

Which, from what you're saying, seems to simply be because where it serves no useful purpose, it tends not to become an issue — not because it has been legislated against. 

So, again I ask: what are you actually proposing? Wanting to remove partisanship is a valid goal, but without something more behind it it's not a lot better than saying something like "We will abolish unkindness". You need to give at least some detail on what a replacement would actually look like, and what sort of legislation could be introduced to get there.

Do you have any examples, from anywhere in the world, of the sort of thing you're suggesting?

philipjohn:
Do parties appear on ballot papers for parish council candidates?

tmtmtmtm:
> Do parties appear on ballot papers for parish council candidates?

They can, though from what I can tell they usually don't. See, for an example, http://www.chelmsford.gov.uk/parish-council-elections, where one candidate stood as "The Labour Party Candidate" (and perhaps ironically was the only one of the six not elected!)

A couple explicitly listed themselves as Independent; most don't use the Description field at all (which is restricted to a maximum of six words).

The Electoral Commission issues separate "Parish and community council elections: Guidance for candidates" for [Standing as an independent candidate](http://www.electoralcommission.org.uk/__data/assets/pdf_file/0004/141799/Part-2a-Standing-as-and-independent-candidate-P-and-C.pdf) and [Standing as a
party candidate](http://www.electoralcommission.org.uk/__data/assets/pdf_file/0016/141802/Part-2b-Standing-as-a-party-candidate-P-and-C.pdf).

tmtmtmtm:
http://www.staffordbc.gov.uk/live/Documents/Elections/Candidates-Guide---Parish-and-Town.pdf also gives an example of the extra form you have to fill in if you want to stand as a Party candidate.

andrewdwilliams:
It's certainly an odd arrangement.

I don't see how having non-aligned candidates are any different to Independents.

I think I might email the Local Government Association to ask them exactly why there is a trend for Town Councils to be party political and Parish Councils to not be, although there seems to be no real basis in law for this.

reducedhackers:
I am dropping in because James asked me for a comment. 

 I am a Councillor for North Horsham Parish Councillor ; having been in this seat for near on 8 years come May . Parties involvement in council meetings ( from my own experience ) can and do occur although not nearly as overtly as can be seen in other levels.  You still see votes that might align by party lines and I have experienced the 'discussions' before meetings where despite the lack of party whip there are opportunities for members to be encouraged to vote along certain lines.

 On the whole the NHPC has tried to operate as a group of independent councillors not as a political party and this is reflected in campaigning come the Elections where candidates for the Parish will not use the local Party to aid their campaigns. 

 Having been involved in two of the local parties; with a view to gaining an understanding of their processes and their makeup as well as assessing if their long term goals are validated by their contributions I can only point out that many factors of a party mechanism exist as artifacts of an era when communication was not instantaneous.

 In effect political party manifestations are the result of habitual practice they are certainly not required in a debate however they seem to benefit quite a few individuals who feel the need for definition in ideas. 

 Meanwhile I have experienced just as much corrosive and divisive moments in debates created by individuals and officers all of whom have had nothing to do with a party and whose contributions are based merely on their own prejudices.  

to try and recap the core issues. 

 * Agenda items can be manifestly controlled by the will of an officer; the will of an interested individual or the agenda or a party all of which can be either beneficial or detrimental to a process. 
 * Political grouping are not  easily defined within the context of an issue; assuming something is more left or right or centre before a meeting is equally a matter of perspective and result. 
 *  Expecting to remove party affiliations from a meeting is a problem as complex as trying to solve electronic voting.   

I hope this input helps.

as per disclosure : 
Im writing this as a member of the public and these views do not reflect North Horsham Parish Council I am however a member of that Council.

andrewdwilliams:
So, in essence, the idea behind the move is sound. Unfortunately, it's hard to pinpoint exactly what needs changing and how exactly to change it.

tmtmtmtm:
@reducedhackers Thanks for the input — that's all very interesting.

tmtmtmtm:
I'm :-1: on this — partially because there's no concrete proposal yet, but more generally because I fear that this strays perilously close to a restriction on Freedom of Association. If people want to group together in political factions they should be allowed to. If anything, having the resources of a larger party to properly research and advise on issues should be a good thing.The fact that sometimes this is misused, and/or leads to outcomes that are less than ideal, isn't a valid reason to remove that ability.

PaulJRobinson:
I like this proposal, but think it would be very difficult to implement at any level of political/democratic institution because you'll always get blocs of like-minded people grouping together, voting together, and organising themselves in order to be productive and to achieve a programme of activity. Unless you were to ban Councllors from being members of political parties (which I think is an infringement of personal liberty) allegiances and groupings (whether formally defined or designated as such) will always be the result, and will always tend to occur between people who happen to be in the same party.

Floppy:
So, in summary, this is an attractive idea, but implementation is really hard without infringing liberties, and because the existing situation at parish level doesn't seem to have any actual backing in law. I've pointed Peter MacFadyen at this conversation in case he has any thoughts to add, though, so we'll see if we can move on our thinking a bit.

Floppy:
Paging @PeterMacfadyen for his comments :)

PeterMacfadyen:
What an interesting stream.....  Parish and Town councils do indeed have the same powers and responsibilities, although some of the latter have now started to enact ‘general powers of competence’ from the Localism Act which gives them the power to anything legal (like start a Housing Company to make profit....as opposed to cutting the grass in the park).  It is also the cases that while many Town Councils have political party based councillors, few Parishes do BUT many Parish councillors also have overt political allegiances, and may stand as party politicians at other (District or County) levels.....  And no, you can’t stop anyone describing themselves however they like as they seek election or after!
For me the political party issue is key because it can often disguise motives.  Why do people stand in these roles?  To whom do they answer once elected?  Standing for a party allows a candidate not to expose other motivations (perhaps around a single issue)..... the bottom line is that it is not helpful to the electorate at these lower levels, as virtually none of the manifesto, or ideological determinants of the party will be relevant at this level.
But it would be naive to pretend that groups of independents do not also behave in tribal fashion – perhaps as a loose grouping of older men who have been councillors for many years; or a new group of ‘incomers’ intent on change.....
In Frome what we arrived at is a ‘Way of Working’ – essentially a set of behaviour, based in part on Martin Bell’s rules devised to expose and dispose of corruption.  These have served us well for 4 years – not always totally adhered to, but as a background guidance enabling us to focus solely on the job in hand.

Floppy:
We don't seem to have reached a workable proposal here, so while the idea is good, I'm closing the change as dead. We can revisit in the future if we can come up with a proposal that resolves the various problems and is actually enforceable.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/283), and results are set in the merge status checks below.

## Changes

PR #285, opened 2015-02-21
Reform the HoL Reform policy

philipjohn:
Adds a commitment to imcremental reform of the House of Lords, given the citizen-panel based reform is quite a big ask and won't happen immediately. End goal copied straight from the previous page.

Also restructures to give HoL reform it's own page for neatness.

### DIFF
--- a/manifesto/democracy.md
+++ b/manifesto/democracy.md
@@ -9,6 +9,10 @@ published: true
 
 We believe that the UK should be a truly democratic country, where the state is properly accountable to its citizens. In the short term, this means improving the existing system in various ways, such as by improving the electoral system, changing how parties are funded, curbing the influence of the lobbying industry, and replacing the House of Lords with an elected chamber. In the long term, we believe that the UK should be a secular republic rather than a monarchy, and that at some point in the future technology will allow us to make decisions using some form of direct democracy rather than needing to elect representatives.
 
+See also;
+
+* [Reforming the House of Lords](house-of-lords.html)
+
 ## Written Constitution
 
 The British constitution should be collected into a single written document that defines and regulates the powers of the British government, and the rights and duties of its citizens[^1].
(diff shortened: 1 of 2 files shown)


tmtmtmtm:
> No ministerial should be held by an unelected politician

Aside from the missing word, this is slightly ambiguous. I'm assuming this should simply say that Ministers can only be appointed from sitting Members of the House of Commons?

I'm curious, though, as to what this proposal would also mean for the Leader of the House of Lords.

philipjohn:
Thanks for the feedback - I've clarified that sentence about ministerial appointments.

> I'm curious, though, as to what this proposal would also mean for the Leader of the House of Lords.

I reckon we're okay on that one, as the policy only stops *new* appointments to the House of Lords (i.e. new Lords). Happy to make that clearer if you think it needs to be?

tmtmtmtm:
I meant that the Leader of the House of Lords is currently a ministerial position, as the point of the role is to represent the government's interest in the House. Abolishing that connection seems like it could be problematic.

PaulJRobinson:
:+1:

Floppy:
@tmtmtmtm I think that implementation of the plan would work out a solution to that problem, so I'm not opposed on those grounds. :+1:

I am a little concerned that this will give undue prominence to Lords reform in the site structure, but that's not a policy issue, so I'll upvote this, and merge when I have a minute to work on the structure after merge to make sure it still works.

philipjohn:
@tmtmtmtm Any chance of a vote from you on this one?

Floppy:
I'm in the process of merging it, but I'm going to split the democracy page in two a slightly different way, if thats OK. I'll take care of it during the merge though.

Floppy:
OK, all done. We now have an "elections" page and a "democracy" page, which balances out the massive democracy page a bit.

philipjohn:
Yeah that's cool

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/285), and results are set in the merge status checks below.

## Changes

PR #287, opened 2015-03-04
Limit funding to individuals and introduce an absolute annual cap.

GarethShapiro:


tmtmtmtm:
Why?

GarethShapiro:
Limit funding to individuals as it is individuals who vote.  Currently people are less influential than companies are, this seeks to give people priority. Even campaign groups such as Greenpeace are subject to funding constraints and this shapes their activities.  Political parties specifically should all be subject to the same funding constraints which result in funds that are directly proportional to the amount of individuals that can be convinced to contribute.

An absolute cap to level the funding playing field.  Ideas should win votes, not cash.  If the ability to influence people is determined simply by the amount of money, or in kind favours, a political party can aquire as it is now then over time the party with the richest donors will simply be able to influence the most people.  In this system there is no way to abstract politics away from finance.

Floppy:
:+1: on the principle of limiting political donations to individuals. This will make parties much more interested in people than in rich individuals or corporate donors. I think there is some stuff in Rebooting Democracy about where this is done elsewhere, I'm just trying to find the reference.

Floppy:
Actually that's about public funding, which is different. There's a good argument to top up individual donations with public funding, but that can be made separately. The principle of getting big and corporate money out of politics is the important thing here.

tmtmtmtm:
Do you have more information on the current constraints on Greenpeace's funding? My understanding was that they choose which donations to accept or not — rather than this being something imposed upon them by law. 

Mega-organisations like Greenpeace, Amnesty etc are also generally the exception, here, in that they have already built up a significant enough body of supporters to enable them to rely on individual donations. I suspect that this proposal would instantly wipe out over 90% of UK Charities that engage in any degree of campaigning.

Floppy:
Ah, interesting. Is the wording intended to hit charities as well? I would suggest it's simpler initially to just keep it to political parties. Campaign and lobbying group cash is perhaps better dealt with through changes to lobbying registers, etc.

tmtmtmtm:
As worded, yes, this quite clearly hits charities etc, too. 

And whilst I have a lot of sympathy for the "removing corporate money from politics" rhetoric, I've yet to see a version of it that actually *works* (outside of totalitarian approaches), as the money can (and will) always simply move back a level from the restrictions (e.g. politically sensitive, but officially non-partisan, issue-related campaigning).

As such I'd *strongly* prefer things like this to be expressed positively, rather than negatively. Let's talk about how political parties _will_ be funded — not how they won't.

GarethShapiro:
It's not intended to hit charities however lobbying is a double edged sword.  I am sure we all welcome lobbying for ideas we agree with but are opposed to it for ideas which we don't.  There does seem to be a massive skew though in favour of lobbyists with the most cash.  Fracking vs Right To Die, for eg.

A personal budget which can only be spent once per calendar year and spread out across all types of political activity seeks to balance the resources that large corporate campaigners are able to accumulate and those realised by much smaller organisations.

It's not perfect as of course there is a disparity in amount individuals can contribute in the same way that business can, I can't afford to give £5k away a year, but my thinking is designed to make each contribution the same value and attempt to remove the massive advantage gained by single large business donors.  A rich person can of course get his rich partner do also donate their budget but a separate process needs to be undertaken each time.

Perhaps the problem is with lobbying and this should be separated from political party funding but it's easy to see how this could simply end up being funding though another channel as are posh dinners (Tory Party) or business breakfasts (Lambeth Council) which my suggestion is seeking to eliminate.

I don't have any specific details on Greenpeace but understood it in terms of these Mega-orgs having substantial costs and due to their success have scaled to a degree where individual donors are not enough to sustain them.

Floppy:
I agree, but that's a massive conversation with many options. Starting with what we want to change about the current system is a good start, so I back this (with clarification about the wording that would hit charities) as a step in the right direction. I've got ideas on funding that I'll happily add immediately as an addition to this once it's in.

Floppy:
@GarethShapiro yeah, I think this might be best kept to direct funding of parties for now. As with the public funding conversation, making lobbying more transparent is a big issue, and I'd love to see the basic idea of changing direct party funding get in to set a direction of travel.

GarethShapiro:
:+1:

Floppy:
You can edit the wording in the files changed tab, if you want to. But maybe you know that, in which case I apologise for stating the obvious. :)

mikera:
:-1: because political parties should be accountable to the public, not donors (of any form). Much better to have public funding of parties based on the number of votes received.

I think it may be acceptable to allow an exception for new parties however (to allow startup costs before their first election)

Floppy:
@mikera I think you might be right, but that's a massive can of worms. I'm entirely happy to open that can, and we should, but as mentioned above, this seems a step in the right direction by removing *some* donations. Would you accept it as a step in the right direction, which we then build upon?

mikera:
@Floppy sure as long as the more advanced idea gets in! It may be a can of worms but if we aren't able to tackle the important issues, who will?

More generally, I think we need to be relatively more bold with our proposals. To catch people's attention, the manifesto needs big, bold ideas rather than "tweaks around the edges". Just IMHO of course but my observation is that few people read manifestos in detail anyway, so the big ideas really need to stand out.

GarethShapiro:
Have we agreed then that party funding should be limited to individuals on the electoral register?  Campaign and lobby groups are to be removed from the wording I suggested.

I am still quite keen on the personal limit, despite the obvious reaction to this being, as tmtmtmtm has stated, the money will move somewhere else.  This to me this reinforces the encapsulation of party and campaign funding allowing separate reforms.

Perhaps another way of addressing the personal limit is to introduce a tax on the individual, rather than the party, once an annual personal limit has been reached.  Individuals can still donate as much as they like but after an amount, obtainable to the average person, has been reached it becomes more more expensive to do so.  This also perhaps allows for the regulation of proxy donations, by other individuals, to avoid the limit could viewed as tax avoidance.  This tax could raise funds for public funding of other parties.

I am also quite keen on funding being in an absolute sense.  If money moves from an individual to a party, it's funding.  If you got to wear a tux to a dinner party as a thank you for your cash then great but it counts to your annual total none the less.

Feel free to let me know if this is not the right time / place to be talking about these details, I'm pretty new to this.

mikera:
I created this PR as an alternative:

https://github.com/openpolitics/manifesto/pull/289

Floppy:
@GarethShapiro don't worry, you're doing it right! All constructive debate is good, it sometimes just takes us a little while to work out the details and arrive at consensus!

@mikera I don't think that this is incompatible with your proposal, which is bigger and bolder, but trying to achieve the same goal of taking big money out of politics. Do you really think we can't accept this as a step in the right direction? I think your larger change will open up a lot of debate (which is great), but will take a while to arrive at consensus, and this could be a good start in the interim. Also it would bring in @GarethShapiro as a contributor and he could then vote on your suggestion. :)

mikera:
Sure... happy to have as an interim step.

Floppy:
That's great @mikera - if you could add a :+1: (or :hand: if you don't want to actually upvote) to counteract your block, that would be great :D

tmtmtmtm:
Surely there's no point on changing the vote yet, as we're still waiting for the text to change, which would reset the votes anyway? (And surely voting yes now, under the new rules, carries a significant risk of the current text being passed as is?)

tmtmtmtm:
One area that hasn't been discussed at all here yet is the impact on trade union donations. Last time I saw the figures, Labour get about 50% of their funding that way. If abolishing that is explicitly part of the goal here, then that should be acknowledged (at least in the discussion).

tmtmtmtm:
@GarethShapiro Are you only referring to money in all this? What about other forms of in-kind donations?

GarethShapiro:
I couldn't see a reference to the text changing and the effect on votes but I have made a change.

I was thinking about trade union donations and I think that they should also be abolished.  If people with capital interests are able to shift their financial contributions to campaign groups then people with labour concerns could just as easily change some of their union subscription to a party donation.  Trade unions would retain enough of an advantage of being in a position to influence their members to donate accordingly.

A downside to this is simply that cash wins prizes, for unions as well as parties, and this could be seen as union busting.

On the upside unions may realise a benefit from streamlining their core finances?  Party donations from union members to parties could itself become more flexible, perhaps spread over more than one party.

GarethShapiro:
What about in-kind indeed.  I need to ponder this a bit more before I will have anything reasonable to contribute.

Floppy:
Sorry @GarethShapiro, yes, when a change is made, the votes are reset, so that things don't change without voting. Anyway, I'm happy with this as it stands, so it's a :+1: from me. Thanks!

tmtmtmtm:
Only the first paragraph has been changed — the second still restricts funding to "campaign, lobbying and interest groups" as well.

I'm still also confused as to what "This…includes every type of fundraising activity in addition to straight forward donations" actually means. As written it seems to say that people can only donate a certain amount of their _time_ to political party activity too, which I suspect isn't what's meant — or if it is, should be clearer (and discussed!)

[Would be good to change the "Only individual's appearing on" to "individuals" too]

Floppy:
Good point there @tmtmtmtm, I missed that it was in the second paragraph as well. :hand: for now, until @GarethShapiro can address those points.

GarethShapiro:
Sorry about the oversight.

It seems the procedure is to keep changing the text and keep voting rather than have a long discussion to get the text right and then change everything, is this correct?

My wish is to reduce the amount of money involved across the whole system with a shift of emphasis towards actual human beings.  This implies that parties need to get people to donate more time than they might currently.  People have to actually get out and about and engaged rather than just sending a cheque off.  The idea here is to attempt to pressurise the monopoly of bigger parties and promote more diversity though smaller parties also being able to afford the currency of campaigning.

Walking about putting leaflets through doors is not fundraising and not a service many people provide for a fee but professional pro bono services should be counted against the providers annual allowance.  Parties are able to call on the resources of their members but there is a limit to how much of an advantage parties with members who have access to greater resources are able to leaverage.

Ultimately my completely unrealistic goal is to reduce campaigning to people on the streets shaking and hands and giving speeches in the community.  While idealistic I think it's worthwhile to aim for that and see how far the idea can go.  More people in the world are able to engage in campaigning activities than there are qualified to research policy.  Parties should be forced through austerity to focus their reduced funding on policy and administration than buying votes through activities such as advertising.

Should I be suggesting changes to the wording based on my ideas here or just changing the text and waiting for a vote?

tmtmtmtm:
I assume some people will be hitting their annual limit of being able to edit this manifesto soon, then?

GarethShapiro:
Oh that's what type of forum this is.

Floppy:
@GarethShapiro it's up to you really - if you want to propose a change to the wording directly, then that's great, and it will reset the votes for another round. If new wording arises out of the conversation first, then that's also fine, but still the edit will trigger a vote reset. It's happened both ways in the past.

@tmtmtmtm I think @GarethShapiro was clear that his change isn't intended to cover time spent, just financial transactions.

From the conversation above, I think if we remove the words "campaign, lobbying and interest groups" from the second paragraph, and perhaps change "fundraising" to the more explicit "financial" in the last sentence, then this should cover it.

@GarethShapiro, this is a good discussion, thanks for continuing to engage. I know the process isn't always easy! I'm keen to get the principles merged in, so let's land this one :)

Oh, one more thing, regarding unions, yes, this would cover that, and I agree with @GarethShapiro that it's just as desirable as getting corporate money out. Force the system to engage with people, not organisations.

Floppy:
Regarding in-kind donations (forgot to address that) the Electoral Commission already has rules around what counts and what doesn't. Roughly, if you're doing your job for free, it counts as a financial donation and has to be reported.

GarethShapiro:
:+1:

Floppy:
:+1:

tmtmtmtm:
Happy to defer this part to another change, especially as I know that at one level the answer is "the pre-existing text", but where does the figure of £5000 come from? At the minute that feels like a number just pulled out of the air.

Is it worth adding anything about independents (or even candidates separately from their parties)? Having it entirely about parties seems to leave a very large gap — but perhaps that covered by other rules around funding of individuals? Other countries have separate limits for these (e.g. [Canada](https://www.opensecrets.org/overview/limits.php)).

Floppy:
@tmtmtmtm yes, it was a change in the early days: https://github.com/openpolitics/manifesto/pull/20. Certainly could be revisited separately, as I suspect it is just pulled out of the air. As for independents, yes, the rule could cover that as well, I think, though it's much less of a problem than big money flowing to parties. Could be added here if @GarethShapiro wants, or we could do it separately.

GarethShapiro:
Happy to include independents here, I hadn't really thought it about it but it makes sense.

I did originally consider changing it to £500 to be honest.

Floppy:
Minimum open period is 7 days anyway, so there's time before we'd be allowed to merge if you do want to add independents, otherwise we can do it later.

PaulJRobinson:
:+1:

Floppy:
For some reason the votes aren't counting right here; perhaps this comment will ping it into life. Anyway, this change is ready to merge, with the required votes, so it's going in. @mikera's other funding change will be ready tomorrow, and we'll have to work out how combining the two works. This is a situation we've not had yet, so it's time to learn :)

Floppy:
Another ping to kick the votebot.

Floppy:
This is passed, so I'm merging. I'll have to fix the robot later.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/287), and results are set in the merge status checks below.

## Changes

PR #288, opened 2015-03-04
Introduced Community Engagement into Society.  Added points about weekend surgeries and hustings.

GarethShapiro:


GarethShapiro:
Many, I would even suspect most, people cannot attend a surgery which runs from 6pm - 7pm on a weeknight.  

Hustings are a traditional and, if you consider the televised version Question Time, popular format whereby the public can scrutinise the thoughts and attitudes of those they elect.  Ward councillors should be prepared to talk in public regularly about their activities, in the company of their contemporaries.  This is also a very efficient way for the public to understand how their own issues and ideas sit relative to others' in the community and also to hold what could be seen as a "Group Surgery" reaching more people in less time.

Floppy:
:+1: nice ideas both

Floppy:
Only concern is that currently there's no duty on anyone to run hustings, even during election time, but that is being proposed as an addition in #272 already, so that's fine.

tmtmtmtm:
Maybe my understanding of "hustings" is off slightly, but the explanation given in the comment above is different from what I expected from the text. I support the general idea, but would like to see it expanded there.

Floppy:
The electoral commission guidance states: "A hustings event is a meeting where election candidates or
parties debate policies and answer questions from the audience"

http://www.electoralcommission.org.uk/__data/assets/pdf_file/0007/105946/sp-hustings-rp-npc-ca.pdf

Does that make sense with what you read/understood?

tmtmtmtm:
Yes — i.e. that it's in the context of an election. Which isn't what's being proposed here.

Floppy:
So I read it as that hustings shouldn't be only during election time, and that regular debates with all local parties should be held. Perhaps @GarethShapiro can clarify his meaning.

tmtmtmtm:
@Floppy — that's what I took the original proposal to be, but the clarification here made it sound like this was more of an event for the elected councillors to talk about what they're doing, face questioning etc

Floppy:
Ah, I see, yes. Let's get @GarethShapiro to clarify which he means, and if it's not a general all-party debate thing, then we should change the word "hustings" to something else.

GarethShapiro:
I specifically meant that these public meeting should be all the time, not just in the run up to elections.  I've used the wrong term "hunting" which needs to be changed. 

Any suggestions?

"public debates"
"public forum"

tmtmtmtm:
I think one of the key questions is who gets to participate (in terms of presenting / answering questions). Is this for the elected representatives? Or including people who ran as candidates and lost, etc?

GarethShapiro:
Chair
-------

I recently went to one locally which was specifically focussed on green issues.  It was chaired by a C class executive of a well known NGO.  While clearly in favour of the better green policies he was very otherwise very good at keeping the whole thing on track and the audience engaged.

It way be a bit much to expect good chairpeople (is that a word?) in every community but then again is this not patronising?  Not sure but this is probably key to how popular these things might be.

Who are our community elders these days?

Panel
--------

An elected councillor and one member from each of the other parties who lives locally.

PaulJRobinson:
:+1:

Floppy:
:+1:

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/288), and results are set in the merge status checks below.

## Changes

PR #289, opened 2015-03-05
Introduce public funding for political parties

mikera:
Money has proved to be a major corrupting influence in politics. This has been notably highlighted in recent years by Lawrence Lessig (see e.g. http://www.ted.com/talks/lawrence_lessig_we_the_people_and_the_republic_we_must_reclaim#t-705558 )

The simplest and most effective way to remove this corrosive influence is to ensure that all parties are funded by the public, and in the public interest.


Floppy:
Thanks @mikera! Is there a reference you could provide of a place where this system is already successfully used somewhere in the world?

mikera:
There appears to be a relatively similar precedent in Australia:
http://en.wikipedia.org/wiki/Political_funding_in_Australia

However they don't seem to prohibit other donations, which I believe is an important part of making the policy a success!

Floppy:
Worth looking at France too, there is something similar there, though I don't have the ref right now.

Floppy:
http://blogs.lse.ac.uk/generalelection/reforming-party-funding-in-uk-elections popped onto my radar today. One of the recommendations it talks about is matched individual and state funding. Basically, the state funds a party to the same amount as it can raise from individuals.

PaulJRobinson:
Very much :+1:

mikera:
@Floppy Matched funding is a truly terrible idea! From a economic perspective, it's effectively subsidising cash donations, which makes the problem even worse!

Floppy:
Fair enough, though in my mind at least it puts the allocation of those funds into the hands of the people. Anyway, this isn't about that idea. I will address the one at hand :)

My concern with the idea of funding being based on the votes at the last election is that it entrenches existing power. 

You do allow new parties to enter and raise money by themselves for a time, but then it puts a ceiling on them once they get to a certain popularity. If you took the Greens as an example, they would no longer be able to raise funds because they have an MP, putting them in the same boat as the Conservatives who have massively more votes and so will get massively more funding. 

Doesn't this stop small parties from growing and challenging the big ones?

Another question: "most recent election". Is that at a particular level, or at any level? Council elections have a much lower turnout, so the funding would be rather... spiky.

I'm a :hand: on this for now; still working through it in my head and checking a couple of references (though I am in favour of the principle of reforming party finance, 100%).

Floppy:
The [Electoral Reform Society's Deal or No Deal report](http://www.electoral-reform.org.uk/downloadfile.php?PublicationFile=446) is very interesting, and recommends three things. Limiting individual donations, state funding of the type proposed here, and a lower spending cap. They also have a number of references to places elsewhere in Europe where similar systems are used.

I'm happy with a hybrid system like that, but as far as I can tell, banning individual donations altogether will put a really ceiling on small parties just starting to break in.

I'd like to work the Green Party numbers to check my assumptions. I'm probably completely wrong.

Floppy:
So, in the last General Election, the Greens got 265,243 votes, giving them funding of £530,846 per year under this proposal.

In their various election returns (which doesn't include non-campaign spending the rest of the time, salaries, etc), they spent:

| Election | Spend |
|----|----|
| EU 2014 | £575,383.19 |
| Welsh Assembly 2011 | £56,800.82 |
| NI Assembly 2011 | £5,822.25 |
| Scottish Parliament 2011 | £132,463.84 |
| General Election 2010 | £353,550.21 |

Not sure if the 2010 spend includes deposits, but that would have cost them around £150k as well.

Their annual accounts for 2013 as reported to the Electoral Commission showed, for the central party only, in England and Wales, income of £881,819.00, and spending of £855,044.00. And that wasn't an election year.

I'd suggest that the Greens are showing that that's not enough money to be effective, and that therefore this proposal puts a difficult ceiling on small parties.

Solutions could be:

* Move the limit up beyond which individual donations aren't accepted
* Remove the individual donation prohibition
* Increase the amount per vote

Floppy:
On a more general note though... can you elaborate on your objection to low-level individual fundraising? 

Most of the objections I've heard about party funding are about the large-scale, corporate and organisational giving, or huge donations by rich individuals. If each individual can only donate £5000 as discussed in other PRs and recommended by the ERS, doesn't that force the funding to be spread out to the people anyway?

How does widely-spread individual donation corrupt politics, other than that parties will listen to their millions of donors?

I agree that parties will always listen to their funders, but below a certain max funding level, individual voices won't be able to stand out from the crowd and gain undue influence.

(again, to reiterate, I really want party funding reform, and I'm working through these thoughts for myself as much as for anyone else, so this is all intended constructively)

Floppy:
Just realised this was also proposed under #194, and some discussion took place there as well. That one will be closed as out of date now, but the discussion might be useful to refer back to.

Floppy:
To summarise my position on this one again; I'm entirely happy with the ideas around state funding here. We could refine them a bit, but that can come later. 

The bit I'm not sure about yet is the complete ban on donations, because we have a 5k cap from individuals only, which seems to me to avoid the usual big money problems.

I'd be entirely happy to accept the change without the ban on individual donations; perhaps we could treat that separately?

philipjohn:
:hand: what @Floppy said :)

mikera:
I'm still believe it is important to end individual donations for many reasons:
- It will inevitably cause parties to spend campaigning money and effort on engaging more with wealthier segments of the population
- I'd much rather see "fundraising" be about engaging and attracting *voters*, especially from under-represented communities 
- No matter how low the cap, there will be *some* influence of money if financial donations are allowed. We want parties to be all about having the right policies, not pandering to meneyed interests in any form
- It sends the wrong message (that politics is about money)
- In terms of donations, I'd much rather see people donate time and energy to their party, offer their home for meetings etc. I have no problems with "in kind" donations of this sort, that is what we want to encourage

To address the issues of insufficient funding for small parties, it may make sense to have a "tiered" model of funding, e.g.
- £10 for the first 1,000,000 voters
- £1 for every voter after that

I suggest we accept this PR, and if necessary refine the exact funding formula later?

iconoclast86:
:hand:

yellowgopher:
Are we talking about central party funding here?  I presume we are.  The problem is that most parties are made up of a multitude of various constituency, interest and other groups and organisations; some of which could be in remit, others would effectively be invisible to the state.  The issue is that large scale donations would (and probably are) being cleverly redirected and the established parties will be able to "manage" any limits and enhance any state funding as they see fit.  That's not to say I am against any state funding but I think if we move towards a democratic system that more reflects the support each party has this becomes less of an issue.  At the moment there are only two UK wide parties that could form the bulk of the next government so the majority of support will go to either of those - regardless of state support.  Make each vote count more then things can change...

iconoclast86:
My idea for funding parties is to give each registered voter a fixed sum to donate to parties. This sum could be donated whole or divided between multiple parties. I envisage this taking the form of an electoral voucher, which parties would redeem from the treasury. For the sake of argument, lets say the voucher would be worth £10 per voter, per annum. Combined with a draconian upper limit on donations, parties would be forced to compete for donations from the general public, This will make political parties more representative.

mikera:
@iconoclast86 your idea is very similar to public funding per vote (i.e. the idea I am proposing)

I think public funding per vote is better however for at least three reasons
a) It is extremely cheap to administer - you don't need to run a complex "voucher" scheme
b) It has less risk of compromising the anonymity of the vote (if you can use the vouchers to track which person has donated to which party, it gives a pretty strong indication of who that person intends to vote for)
c) It reduces the incentive to actually vote (people may send in their voucher but still not bother to vote when they are a minority party)

yellowgopher:
I agree, public funding per vote is easier/cheaper to implement but I like the idea of a voucher system.  The anonymity issue can be overcome with careful design - in fact you could probably trace who people vote for now if you are clever and quick enough - we are all issued with unique numbered voting slips!  The incentive side is a problem but I'd argue it would be down to the parties themselves to mobilise their voters - at least they would have an idea of who would be voting for them by the number of vouchers received (how about that for accurate polling?).  £ per vote wouldn't help new parties that have never been part of any poll - would we consider start-up funding for these?

mikera:
@yellowgopher I agree it is important to address the issue of funding new parties before they have enough votes - that's why this PR has a clause allowing campaign donations for small / new parties until they are established.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/289), and results are set in the merge status checks below.

## Changes

PR #309, opened 2015-04-01
Education funding changes

yellowgopher:
Split from previous proposed update.  

Move towards an individual pupil funding approach.

### DIFF
--- a/manifesto/education.md
+++ b/manifesto/education.md
@@ -15,7 +15,11 @@ We believe that learning is life long and should be encouraged in all elements o
 
 ## Funding
 
-Education should be freely available to all to first degree level or equivalent. Tuition fees for university students should be abolished, as university education is of general benefit to society, and should be covered by general taxation.
+Education should be freely available to all to first degree level or equivalent. 
+
+Tuition fees for university students should be abolished, as university education is of general benefit to society, and should be covered by general taxation.
+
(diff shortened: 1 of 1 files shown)


philipjohn:
This sounds good. I have a concern about how quick "the system" will be able to respond to changes in individual circumstances, and the admin overhead such a system will require. That's all implementation detail though - this vision sounds good, so I support it.

For now I'm a :hand: though. You'll get my vote if this bit goes ;) : 
> These funds will be valid as payment towards private education.

yellowgopher:
So why against using these funds for private education/tuition/support?  Many people already hire private tutors to support study etc?

philipjohn:
Private study as an *addition* to a state-funded, secular education is entirely acceptable, if parents wish to do that, but that's individual choice and they can fund it themselves.

PaulJRobinson:
"These funds will be valid towards private education". So is this a proposal for so called 'school vouchers'? i.e. parents can take the money spent on their children's education by the state and use that at an independent school of their choice, and then top it up with their own cash? Because that won't actually help poorer children get into independent schools. It will just inflate the fees from what they are now, to 'plus the addition of the voucher'.

yellowgopher:
There is the argument that private school fees are paid in addition to taxes that fund state school education (there is no rebate if state schools are not used but the funding is there and can be used for other children).  And, as for the inflation of fees for private schools, private schools already do a lot of work to reduce fees for poorer children, have to maintain a charity status by proving they are working for charitable aims and it may even result in an increase in private provision.  However, on balance, I will remove this to get the change through.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/309), and results are set in the merge status checks below.

## Changes

PR #310, opened 2015-04-01
Local schools

yellowgopher:
Replacing Secular Schools policy with Local Schools.
The problem with actively promoting the removal of Faith Schools is that there is a broad section of society that see value in Faith Schools and religious direction.  If this is to be a truly "people's manifesto" we need to work on mechanisms to promote more local influence in how educational establishments are set up and run; within a basic framework specified by the state.  Otherwise we risk the development of two distinct systems in the country - something that is happening to a certain extent right now with a Maintained and Academy/Free school programme running side by side.
Allowing for more local influence will naturally direct educational establishments towards the general consensus of the people - if, for example, the vast majority of local people do not have a religious belief then educational establishments should move in that direction (if governors/trustees are properly representative).
Closing down the last vestiges of Local Authorities and focusing on enhancing the Academies/Free School programme (and converting remaining Maintained schools) so that it does provide what is required in local communities reduces complexity and enables the state to focus on developing one system.

### DIFF
--- a/manifesto/education.md
+++ b/manifesto/education.md
@@ -17,26 +17,19 @@ We believe that learning is life long and should be encouraged in all elements o
 
 Education should be freely available to all to first degree level or equivalent. Tuition fees for university students should be abolished, as university education is of general benefit to society, and should be covered by general taxation.
 
-## Secular Schools
+## Local Schools
 
-50% of the British public identify as having no religion, and this number is growing[^1] but over half of all state funded schools have a religious character[^2], including over 4,500 Church of England schools. It has been demonstrated that faith schools are religiously selective, excluding those of other or no faith[^3].
+Schools should reflect the local aspirations and desires of the communities they serve. Mechanisms will be put in place to allow local communities more say in how schools are set up and run - including, but not limited to:
 
-Therefore, we will mandate that all state-funded schools be secular in nature.
-
-No state funded school will be permitted to admit students based on religiosity.
-
-We will end compulsory worship in schools.
-
-State-funded faith schools will also lose their exemption to equalities legislation, preventing them from practising religious discrimination in their employment policies.
-
-Religious education (R.E.) will also be reviewed, starting with research on the current state of R.E. and followed by a consultation to include all interested parties on whether to keep R.E., reform it, or replace it.
-
-Scottish authorities will no longer be legally obligated to appoint unelected religious representatives to school boards. Instead, elected councillors will be appointed by a vote of the full council.
+- powers to promote a broad cross-section of local people sitting on governing bodies.
+- legal requirement to consult on progress with local communities at set intervals.
+- further evolution of the Free School and Academy Programme to facilitate the setup or conversion of schools of any faith, speciality or none.
+- An end to a two tier approach to school management - all maintained schools will be converted to Academies and current Local Education Authorities closed but a framework will be created to allow new style Education Authorities to be set up to manage chains/groups of Academies and Free Schools should local communities desire this.
+- Inherent in all the above will be a desire to protect smaller local and rural schools and their individual characters. We understand that this will come at substantial cost to the taxpayer and certain criteria will have to be met (i.e. school is well performing, has great potential etc) but this fits in with the local aspirations that we wish to encourage.
 
 ## School Governance
 
 All state secondary schools will have an elected students council, and student president. The student president, who will need to be at key stage 4, will represent students on the board of governors.
 
-[^1]: [British Social Attitudes Survey](http://ir2.flife.de/data/natcen-social-research/igb_html/pdf/chapters/BSA28_12Religion.pdf) (pdf)
-[^2]: [Maintained Faith Schools, Department for Education](https://www.gov.uk/government/publications/maintained-faith-schools/maintained-faith-schools)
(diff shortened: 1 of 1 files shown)


philipjohn:
This unfortunately sounds very similar to what we have already - the power for anyone, with any vested interest to set up discriminatory schools. As is clear from research into the intake of faith schools at present, they are perpetuating inequality and division in society. For that, this policy goes against our core values IMO, so I'm a :-1:

yellowgopher:
It may go against YOUR core value but I am not sure you can argue it goes against local democracy and the ability for people to choose...?
I'd argue that a purely secular approach is divisive.  I am divided!
If we are moving to a secular society you would expect the majority of schools to be secular in nature under this proposal.  You would expect a gradual tailing off of faith or otherwise directed schools.  This is an easy way to give the power to the people - which is a core value isn't it?

philipjohn:
No, the core values of the manifesto [as stated](http://openpolitics.org.uk/manifesto/). Specifically;

> ...individuals should be.. Free to act in any manner that does not harm another individual, and *does not infringe upon the rights of other individuals*.

The current system of "free schools" that can be set up by "anyone" in the local community has resulted in more and more schools set up by groups with narrow interests. It has seen good, local schools closed against huge local opposition with the land handed over to private groups and out of local government hands - from oversight by locally elected politicians to an unaccountable group of people with the money to get started.

What matters is the *education* children receive and granting schools run to "reflect the local aspirations and desires of the communities they serve" does not equal a robust curriculum built by experts from a foundation of evidence. The "local aspirations" will inevitably mean the aspirations of the local vocal majority, not the whole community.

yellowgopher:
I absolutely agree - the education children receive is most important criteria - and this can partially be acheived by a national curriculum that outlines the core requirements - the Free Schools project doesn't have to exclude this!

However, the argument here is about removing the boundaries of state where possible and allowing groups and organisations to set up schools that reflect (reflect, not adhere religiously to) their aspirations and communities.  It ties in neatly (and can be ring fenced by) the core value:

"...individuals should be.. Free to act in any manner that does not harm another individual, and does not infringe upon the rights of other individuals."

The move to secular schools immediately disenfranchises particular groups - which does infringe upon the rights of other individuals in my humble opinion!

Beyond this argument I have yet to see any real evidence that Free Schools and Academies have resulted any more damage to local educational provision over what was already happening - a lot of the schools "taken over" were failing schools anyway - often with poor infrastructure and demotivated staff.  In such cases there was always going to be opposition but doing nothing shouldn't have been an option either!  As for accountability, the accountability moves right to the top - the DfE - in these cases!  But where were the LAs etc when these schools were failing in the first place?

You do rightly point out the risk that a vocal local majority would set the agenda but work could be done to get a fairer reflection of the community as a whole - but, ultimately, one group will lead.  Under the current manifesto pledges, the secular group leads; where is the "whole community" in that!?!

yellowgopher:
Not that I am one to quote the BBC often but http://www.bbc.co.uk/news/education-32038695
Interesting analysis of the overall performance of Academy chains etc.  Not brilliant but not terrible either.  At least no worse than maintained schools.  And often starting from a different position as well.

I am a governor of an academy and, to be honest, I am not sure I would have voted for academy status if I had been there when it converted.  However, I don't see it as a "bad thing" if the relevant funding and support is in place - and it is evident, from some academy convertions, that they can be very successful.

yellowgopher:
Oh dear.. .another BBC report (I seem to attract them at the moment) but worth mentioning here - Faith Schools are popular and changing their entrance criteria to reflect a more pluralistic society...

http://www.bbc.co.uk/news/education-32587694

Successful, popular, taking into account changes in society, sounds good to me!

Floppy:
Interesting; I'd like to know more about how they intend to handle religious instruction within the school for children whose parents aren't religious, or are non C of E.

yellowgopher:
Well, it is handled in a roundabout way already.  We know that there is a legal requirement to teach RE and that non-statutory guidance advises that RE reflects the denominations and faiths of the country at large.  Ofsted do monitor this and will publicise it in their reports if it isn't seen representative enough.  So it is happening anyway and my experience of C of E or other faith schools (I used to be a governor at a Quaker school) is that they manage it just fine - in the end teachers are professionals and it is fair to say a number do not have a strong faith or any at all, so they do the job they are paid to do.  Now this argument is not to say that there shouldn't be changes to ensure everyone has equal opportunity - wherever they live and whatever their beliefs are - but it does suggest that we are at risk of "throwing the baby out with the bath water" by just denying the existence of schools like these.  Rightly or wrongly (in your or my opinion) we risk setting risky and bigoted precedent in exactly the opposite direction of what may or might have been before.  Best to equalise opportunity, leave it to people to decide and be facilitators, not organisers!

philipjohn:
> a lot of the schools "taken over" were failing schools anyway

This is simply untrue. Case in point, an outstanding school being forced to close and become a faith school: https://www.tes.co.uk/article.aspx?storycode=6007907

yellowgopher:
The example you give is of a council maintained school, not an academy or free school.

It is also the result ofca local council policy to move from first and middle schools to primary schools.

In theory, a proposal to allow local communities to set up and run schools would counter this!

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/310), and results are set in the merge status checks below.

## Changes

PR #323, opened 2015-04-14
Devolve minimum wage policy

digitalWestie:


Floppy:
:+1:

mikera:
I like the idea of devolving the minimum wage policy but this commit contradicts this by saying it should be set to the living wage. I think surely if you devolve that decision you have to let the individual regions set the level, surely?

digitalWestie:
Yes, maybe I should reword the above to be a principal that we support at
the devolved level?

Floppy:
My understanding of the change was that the minimum wage should be the living wage, and that the value should be set on a devolved basis dependent on the local living wage.

digitalWestie:
Yep that's what I was getting at.

Floppy:
Perhaps "devolved" is the wrong word, as it has a specific meaning politically. Perhaps just "set locally"?

digitalWestie:
'devolved parliaments and assemblies' to be specific

yellowgopher:
The issue with "set locally" is that what defines "local" - parish, district, county...?
Within a district you can have massive differences in wealth (and, therefore, cost of living).
I like the idea of a "living wage" but I feel this has to be set nationally or the whole thing becomes a complete nightmare to manage and implement!

philipjohn:
I like this. One sticking point - presumably England's living wage would be set by Westminster but to what? The London wage? Or would would the GLA be given power to set London differently?

iconoclast86:
:hand:

digitalWestie:
Since this is generating a few more questions I've added a bit more clarity. In summary what I mean is this:
- Min. wage should be set at devolved level
- The devolved parliament/assembly level means (Welsh/Scot./NI/London assembly)
- In the context of this manifesto this would include the English regional assemblies

philipjohn:
:+1:

Floppy:
Can we change the text "minimum wage policy should be decided" to "living wage level should be decided"? The policy is universal, minimum==living. You want to set the living wage regionally. Which makes sense.

digitalWestie:
good idea, done

Floppy:
:+1:

philipjohn:
:+1:

mikera:
Two points I'd like to make:
a) There is overlap between this and any proposed "citizens income" policy. If the aim is to ensure a living wage, then presumably the minimum wage should be set at a level that "tops up" the difference between the citizens income and the minimum wage?
b) While evidence is inconclusive, there does at least seem to be some US evidence which suggests that minimum wages cause net job loss among low-skilled workers. see e.g. http://www.nber.org/papers/w20619

I don't really want to vote against as I'm in favour of improving the lot of low wage workers, but I do think this policy still needs some work.

digitalWestie:
I've had a look at the studies about minimum wages causing job losses too. In many senses I would prefer mechanisms that would prevent exploitation without the need for legislation to step in. The Nordics have relatively high wages and yet have no minimum wage laws. This is because the majority of the population belong to trade unions that help workers negotiate pay. I actually think we should be thinking hard about ways to help people negotiate pay and to help them understand their worth as workers. Forcing companies to reveal gender pay differences is good example of this. Also, I think encouraging worker-ownership/control would go a long way to this end too. 

Putting citizen's income to one side, I might go as far to say I would stand by a living wage policy _even if_ it meant job losses. No living wage is economic dishonesty - when people aren't paid enough to live on as this means that the welfare system, or increasingly, family members have to pick up the slack.

Anyway, good point though, do we have living wage policy if we go for citizen's income? My first reaction is that it would mean a lower living wage?

mikera:
@digitalWestie The living wage isn't defined as being "paid enough to live on" however so it is hard to defend it as a minimum on a purely ethical basis. I live in Asia and assure you that the living wage as defined in the UK is actually quite a high income by global standards.

The fundamental economic problem is this:
a) You have an employer willing to hire an unskilled worker for a new role. This employer can pay up to £8 per hour and still be profitable, but beyond that would lose money on the margin. Hence the job is viable at any wage up to £8 (this is the demand curve for labour, if you like)
b) You have a workers who is willing to work for £6 or more per hour (and able to get by on that amount), but having trouble finding a job otherwise.

Without a minimum wage, the job would be created (with a salary somewhere between £6 and £8), and both parties would be better off. I find it hard to describe this as "exploitation" - you have two parties willing and able to enter into a business relationship. With a minimum wage set above £8 (say £10), you have just eliminated the possibility of this job being created.

That's why I prefer a citizens income approach to solve this problem. With a citizens income approach, you can pay a citizens income (say £4 per hour), the job is still created (at £6-8 per hour) and the worker gets a total of £10-12 an hour, which is above your original minimum wage but without the problem of a viable job being destroyed.

Hope that is a useful illustration.

Please don't get me wrong: I want to help the poorest in society and improve the lot of low-wage workers. I just think we should be very careful about jumping to the conclusion that a minimum wage is the right way to do this. It's one of those ideas that sounds good at first but has quite a few unintended consequences. In particular, it runs the risk of slightly helping some of the working poor with a bit more income (which is nice, but could be achieved in other ways), while excluding some others from the labour market entirely (which is extremely damaging).

digitalWestie:
I believe the exploitation is created where choice is limited. If there's no citizen's income then it's not so easy for someone to shop for other job offers, therefore I may have to go with whatever wage is available (hence beginning of exploitation). If there's a citizen income this problem is solved. 

Anyway, with ambiguity here, we should decide on Citizen's Income, @mikera do you want to start a vote on it?

Floppy:
A couple of points here:

1. This PR is not proposing the minimum=living wage; we already have that in. This is proposing the the level should be set regionally. The diff is a little confusing, because of a known bug in the editor. @mikera if you disagree with the minimum-wage concept as a whole, that should go in a separate proposal.
2. We have a bit in about basic income already, but it can definitely be fleshed out, and I'd invite you to do so: http://openpolitics.org.uk/manifesto/economy.html#social-security
3. If we move to say that the citizens' income should be adopted (which I would agree with), then yes, this wording would change to take account of the citizens' income; I would imagine it would say something like the minimum wage would be the living wage minus the citizens' income.

BUT, this PR is just about regional allocation of the minimum. Let's vote on that :)

mikera:
Happy to do a separate PR on citizens income, assuming changes can be cascaded to related stuff (like this).

However this commit needs cleaning up first, there is unrelated stuff in it (like financial transaction tax). So still :hand: until is is cleaned up!

Floppy:
No, that diff is just an artefact of the editor being weird, honestly. @digitalWestie you might be able to fix it by pulling openpolitics:master into this branch and pushing again - can you have a go?

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/323), and results are set in the merge status checks below.

## Changes

PR #324, opened 2015-04-14
remove bishops from house of lords, disestablish the church of England, create a secular state, 

andrewedmondson:


Floppy:
Thanks for the additions! Pretty sure we have these two things in elsewhere; let me find them...

Floppy:
Yep, here we go:

Disestablishment: http://openpolitics.org.uk/manifesto/society.html#secularism

Looks like the bishops wording (which we did have before) has been replaced with a more general reform, but we could have it back in.

andrewedmondson:
Thanks James. No harm in repeting them in overlapping sections. I'm sure 
there are quite a few other changes I could mek. Running out of time now.

Regards

Andrew

Floppy:
True that. We can always cross-link after. :+1:

yellowgopher:
My two penny's worth (although I KNOW I am in the minority here) - I would be very careful about explicitly naming groups or individuals in these things because there is a risk that you could end up missing someone or something off the list!  I am actually for the disestablishment of the CofE (I know, shock horror) as I think it is actually a chain around the CofE's neck and a fundamentally unfair position over the multitude of different groups that exist in the religious sector and beyond.  

It may shock the secular majority here that some bishops are actually worth their position in the Lords - but we'd want to reform the Lords anyway... ;-)

andrewedmondson:
It is valid to name the C of E because it is part of the government.

yellowgopher:
Not sure what you are getting at here.  I would create a general framework for a new Lords - removes the issue of particular groups being over represented etc.  But I think we should allow any capable person to serve - regardless of their opinion on God!

andrewedmondson:
Yes I agree. No reason why a bishop shouldn't be there; just not by right.

yellowgopher:
Fair point and agreed - anyone can serve as long as they get there on merit.

philipjohn:
Not sure I like the idea of duplicating stuff. If it really needs to be in two sections, it should just be linked.

PaulJRobinson:
Get's my vote :+1:

iconoclast86:
:hand:

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/324), and results are set in the merge status checks below.

## Changes

PR #335, opened 2015-04-19
Removing the undue influence of the City of London in Parliament

andrewdwilliams:
This is a policy to try and remove some of the unfair influence that the City of London Corporation has in the everyday goings on in Parliament. I don't see how measures like this can be upheld in a modern age like today. I suppose any other measures to curb the City's influence could come under here too. This is what George Monbiot wrote about the City Remembrancer: "The City of London is the only part of Britain over which parliament has no authority. In one respect at least the Corporation acts as the superior body: it imposes on the House of Commons a figure called the remembrancer: an official lobbyist who sits behind the Speaker’s chair and ensures that, whatever our elected representatives might think, the City’s rights and privileges are protected."

http://www.theguardian.com/commentisfree/2011/oct/31/corporation-london-city-medieval

### DIFF
--- a/manifesto/democracy.md
+++ b/manifesto/democracy.md
@@ -1,7 +1,5 @@
 ---
 title: Democracy
-layout: policy
-published: true
 ---
 
 * table of contents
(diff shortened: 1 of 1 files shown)


PaulJRobinson:
I support this in principle. I may be mistaken but I thought we had already passed a proposal on something very similar quite early on. Will need to check to see how this PR differs from what was passed earlier.

andrewdwilliams:
I think you might have been confused (although I don't know for sure) with Democracy 8.15. This removes the role of City Minister, which is an actual parliamentary role although it appears to do something similar. I was wondering whether it would be better to pull this policy under a more general title like "Lobbying", as this policy is effectively curbing lobbying, just a peculiar strand of lobbying.

iconoclast86:
:hand:

andrewdwilliams:
@iconoclast86 Why the abstention? If you don't agree with the policy tell me which bit so I can fix it?

philipjohn:
Yep, I think we have the City Minister somewhere else - have a feeling I proposed that one.

:+1: to this.

Floppy:
:+1:

Floppy:
The actual change is:

> To remove the undue influence of the City of London, we should remove the City Remembrancer from the floor of the House of Commons and end the Remembrancer's privileges to view legislation during the drafting process. As granting these permissions is under the control of the Speaker, legislation should be enacted to ban the Remembrancer from resuming these permissions.

The rest of the change is editor junk, which is due for a fix after the election calms down.

Floppy:
@andrewdwilliams am I right in thinking this has been replaced with the new one merged in #360?

Floppy:
And in #369 as well.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/335), and results are set in the merge status checks below.

## Changes

PR #345, opened 2015-05-06
egalitarianism 

LeilaCCG:


iconoclast86:
:hand:

yellowgopher:
I think I am ok with this - it's difficult to implement; extremely so... by :+1:

philipjohn:
This is great as a principle, which should perhaps go on the front page but I feel like we should have more specific policy aims here.

How do you propose we can make society more equal?

Floppy:
Agree with including this; in fact, it should definitely be a principle on the front page; equality before the law, and promoted as a goal of society in general. I suggest we accept this and then move it to the front page under another PR. :+1:

PaulJRobinson:
:+1:

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/345), and results are set in the merge status checks below.

## Changes

PR #348, opened 2015-05-06
Expanded religious courts section

andrewdwilliams:
We only really have problems with one type of religious court in this country, so I've expanded the section slightly and included a reference to one of the more prominent anti-Sharia campaigns, from which there is undoubtedly more legislation to draw from in the future.

### DIFF
--- a/manifesto/crime.md
+++ b/manifesto/crime.md
@@ -1,12 +1,8 @@
 ---
 title: "Crime & Justice"
-layout: policy
-published: true
 ---
 * table of contents
 {:toc}
-
-What policies should we adopt to ensure a fair justice system?
 
 ## Capital Punishment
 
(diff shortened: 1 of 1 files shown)


yellowgopher:
I'm a :+1: with this one - when in Rome do as the Romans do.

philipjohn:
We shouldn't limit ourselves to Sharia courts, but I don't think this change does, it just adds them as an example. :+1:

andrewdwilliams:
@Floppy Could this be merged now? With the VoteBot down, I don't think there's any other way to do it than just ask.

philipjohn:
All the checks were passing actually, so I've just merged this :)

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/348), and results are set in the merge status checks below.

## Changes

PR #350, opened 2015-05-06
EU Policies added

andrewdwilliams:
Added one about ensuring our membership is cost neutral - we get out what we put in, not necessarily just straight up money but through investment, development, etc.

Also added one about the possibility of leaving the EU and if we did, joining the EFTA.

### DIFF
--- a/manifesto/foreign_policy.md
+++ b/manifesto/foreign_policy.md
@@ -1,18 +1,18 @@
 ---
 title: Foreign Policy
-layout: policy
-published: true
 ---
 * table of contents
 {:toc}
-
-What policies should we adopt in our relations with other countries?
 
 ## Internationalist Stance
 
 We believe that a clear view of the role that the UN, EU, NATO and other international bodies play in reducing the likelihood of war should be clearly and unambiguosly communicated to the electorate. 
 
-The UK should also use it's membership of such bodies to influence them to reinforce this role and use this peaceful approach whenever possible to avoid conflict.
+The UK should also use its membership of such bodies to influence them to reinforce this role and use this peaceful approach whenever possible to avoid conflict.
+
+## Foreign Aid
+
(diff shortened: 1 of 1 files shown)


mikera:
Can we separate into two separate PRs? They seem like different issues to me, both worthy of a good debate

Floppy:
Yes, I think that would be a good idea.

yellowgopher:
I'm a Euro Sceptic but I think any membership of international organisations like the EU doesn't HAVE to be cost neutral - it should be about sharing ideas and resources for the greater good and, because we have inequality, some people/countries have to accept they might pay more.  Now I can see that "greater good" could equate to a value but I am concerned, as soon as you put a figure on such things, it only serves to turn people against the idea ("it costs THAT much!?!").  

Second idea - if we leave to try and join the EFTA - is sound.

philipjohn:
Cost neutrality sounds like a purely quantitative goal that ignores the actual benefits.

mikera:
I don't think cost neutrality should be an explicit objective, but there I would support something about "all member states sharing fairly in terms of overall benefits and costs". 

If there is to be a redistributive element (i.e. richer nations being greater net contributors) then I think that the level of this redistribution should be explicitly agreed (e.g. 2% of GDP from each country pooled and re-allocated according to population size would have a modest redistributive effect).

If you don't do this, then you end up with a bunch of "back door" redistribution happening through various channels (CAP, Structural fund, Cohesion fund etc.) which is driven by political / bureaucratic agendas isn't really transparent or fair.

andrewdwilliams:
So we agree that joining the EFTA is sound, but cost effective is not. I'll exclude cost effective and edit this to reflect that later today.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/350), and results are set in the merge status checks below.

## Changes

PR #362, opened 2015-05-17
Ban public officials from standing for another elected position

philipjohn:
Think of this as the "Boris law". Anyone who is already holding an elected office should not be seeking election to another position without making a commitment to resign first.

*Public office is not a gravy train.*

### DIFF
--- a/manifesto/democracy.md
+++ b/manifesto/democracy.md
@@ -71,7 +71,7 @@ MP salaries should be linked to other public sector wages, and should rise and f
 
 Serving MPs should not hold second jobs unless they can show that there is no conflict with their role as MP, both in terms of time required and conflict of interest. This includes directorships and advisory posts.
 
-All persons holding public office will be barred from standing as a candidate in any election other than for re-election to the same post.
+All persons holding elected public office will be barred from standing as a candidate in any election other than for re-election to the same post. An exclusion will apply to holding parish and district-level councillor positions simultaneously.
 
 Allowed expenses would be reduced to only those costs incurred as part of their job; for instance, travel and accomodation.
 
--- a/manifesto/democracy.md
+++ b/manifesto/democracy.md
@@ -71,6 +71,8 @@ MP salaries should be linked to other public sector wages, and should rise and f
 
 Serving MPs should not hold second jobs unless they can show that there is no conflict with their role as MP, both in terms of time required and conflict of interest. This includes directorships and advisory posts.
 
+All persons holding public office will be barred from standing as a candidate in any election other than for re-election to the same post.
+
 Allowed expenses would be reduced to only those costs incurred as part of their job; for instance, travel and accomodation.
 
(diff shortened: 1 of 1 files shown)


yellowgopher:
I'm for this; although notfor the "gravy train" reason -  I believe holding a second position would be detrimental (in terms of time, energy, passion etc) to the first, or vice versa. The only things to consider are where these posts are junior - say parish council and governing body for example. Also the wording as it stands could be read as a council clerk couldn't be a governor at a school. If this can be made clearer then it will be a thumbs up from me.

andrewdwilliams:
In essence, I agree, however I do come to an issue when it comes to local government. If you don't allow, for instance, Parish Councillors to also be Borough Councillors, and in some cases also County Councillors, then that extends a problem that we already have in this country - not enough people willing to be councillors. A solution could be that you can only simultaneously be a Councillor on two different local government authorities, and no more than two. But I think by stopping a person for being a Councillor on at least two different bodies will create a supply problem, and that helps no one.

philipjohn:
I read your mind @andrewdwilliams :)

Latest update is clearer on "elected" (thanks @yellowgopher ) and excludes parish and district/borough-level.

andrewdwilliams:
Would upvote but I don't think my upvote-y thing actually works. :+1:

yellowgopher:
Going to be a pain Herr mainly because I have had first hand experience of district councillors who sit at parish level effectively having a conflict of interest over areas such as planning! Supply issue or no supply issue I think they shouldn't stand at both levels. Districts, for example, have power to appoint to Parish if need be anyway.

philipjohn:
Yep, I get that @yellowgopher . Unfortunately I don't think that's going to change until we can change the system of Government at the local level. We did have a couple of proposals around that but neither passed. Until we have that, would you consider voting for this to at least bring it in for the levels of government above the parish/district?

Floppy:
A thought - should we allow them to *stand*, but if elected they must resign their other post triggering another election? I'm in favour in general though.

andrewdwilliams:
I personally don't think that we can do much better than this under the current system with the current level of engagement. If we do either of what @Floppy or @yellowgopher suggest, then we end up with, as I say, a supply problem with not enough people filling Councillor roles. And this ultimately leads to the most vacancies on Parish Councils, as they are the least important, and it effectively makes Parish Councils redundant or just more parochial as there are no elections as there are not enough challengers to seats. It already happens with my local Parish Council, we have two or three vacancies all the time.

philipjohn:
Hmm, I thought about doing that @Floppy but didn't because even standing as a candidate can be quite intense so I wondered whether it would have the same effect as holding two positions.

Floppy:
OK, I'm in. :+1:

philipjohn:
@yellowgopher Potentially, we create a "consitutional conundrum" with the supply issue you mention. It'd be good to revisit the proposals around restructuring local government, so if you fancy having a look at that it'd be good :)

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/362), and results are set in the merge status checks below.

## Changes

PR #363, opened 2015-05-26
Homeopathy

andrewdwilliams:
Stop spending money on homeopathy. It's proven to not work. It's as simple as that.

### DIFF
--- a/manifesto/health.md
+++ b/manifesto/health.md
@@ -43,9 +43,9 @@ Examples;
 
 Increase awareness and education in attempt to remove the stigma surrounding mental health issues, that should be taken just as seriously as physically health issues.
 
-## Homeopathy
+## Review of placebo treatments
 
-End public spending on homeopathy within the NHS.
+We will conduct a science-led review of placebo treatments in the NHS, such as homeopathy, to decide whether placebos are value-for-money, ethical and effective at treating patients.
 
 [^1]: [Impact Assessment: A Minimum Unit Price for Alcohol](https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/157763/ia-minimum-unit-pricing.pdf)
 [^2]: [Bargain Booze: Blackthorn Cider 3ltr](http://www.bargainbooze.co.uk/products/4043/index.html)
--- a/manifesto/health.md
+++ b/manifesto/health.md
@@ -1,12 +1,8 @@
 ---
 title: Health
-layout: policy
-published: true
 ---
 * table of contents 
 {:toc}
-
-How do we ensure a healthy population?
 
 ## Depoliticisation and protection of the National Health Service
 
@@ -42,6 +38,14 @@ Resulting price increases would largely be restricted to off-trade retailers suc
 Examples;
 * A 3 litre bottle of 4.7% ABV cider can be purchased for as little as £3.99[^2] and contains 14.1 units of alcohol. An MUP of 45p would force this up to at least £8.81 (£6.35 MUP + £1.27 VAT + £1.19 duty).
 * A pint of 4% draught beer can be purchased for around £3.21[^3] and contains 2.3 units of alcohol. An MUP of 45p would make the minimum legal price £1.65 (£1.35 MUP + £0.21 VAT + £0.09 duty), well below the current price.
+
+## Focus on mental health in young people
+
+Increase awareness and education in attempt to remove the stigma surrounding mental health issues, that should be taken just as seriously as physically health issues.
+
+## Homeopathy
+
+End public spending on homeopathy within the NHS.
 
 [^1]: [Impact Assessment: A Minimum Unit Price for Alcohol](https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/157763/ia-minimum-unit-pricing.pdf)
 [^2]: [Bargain Booze: Blackthorn Cider 3ltr](http://www.bargainbooze.co.uk/products/4043/index.html)


Floppy:
:+1:

philipjohn:
:+1:

tmtmtmtm:
:hand: I don't remotely support homeopathy in the slightest, but I'd like to see more rationale as to why it's deemed enough of a problem to justify removing the decision-making power from local doctors and NHS trusts. I'd like to see _less_ of that sort of thing, rather than more (even if this happens to be one that I would otherwise support.)

Floppy:
Evidence based medicine. The NHS shouldn't spend public money on fairy dust.

tmtmtmtm:
Why is it central government's position to decide what is acceptable or not, rather than those actually providing care? This seems like a very sharp two-edged sword to me. In general I suspect we get better results all around by pushing those sorts of decisions to the front-line, rather than being in the habit of issuing central decrees (almost always by people who have no actual medical qualifications or experience).

Floppy:
Because central government pays the bill, and should demand that what it's spent on is based on evidence. That's why I'd prefer wider wording on this, because if someone could make the case with evidence that giving someone homeopathy would produce the required result (perhaps psychological or social rather than medical), then I'd be inclined to allow it. However, I'd happily add that as a subsequent change.

tmtmtmtm:
If there was wider wording about evidence based medicine, I'd almost certainly accept it. But singling out this particular issue troubles me, unless there's evidence that it's a particularly egregious case (which doesn't seem to be so). Governments imposing their personal bête noires on the NHS seems like a much worse problem.  

FWIW my understanding is also that homeopathy _can_ be effective purely as a placebo. But I'm not seeing calls to abolish all placebos…

Floppy:
If you want definite evidence of spending on homepathy within the NHS: we do have three entire homepathic hospitals, such as http://www.uhbristol.nhs.uk/patients-and-visitors/your-hospitals/bristol-homeopathic-hospital/. That seems rather odd to me. Are there solid medical reasons for such facilities existing? Note that it includes a "Complementary Cancer Care service": http://www.uhbristol.nhs.uk/patients-and-visitors/your-hospitals/bristol-homeopathic-hospital/what-we-do/

tmtmtmtm:
Yes, it seems odd to me too, and I'd definitely be interested in hearing the rationale behind them. Primarily, though, the question isn't "Is there money being spent on it?" — there's no doubt that there is. It's a more general question as to when something it's a *big enough* problem to justify imposing a specific policy over the wishes of front-line professionals. This happens to be one I'm in favour of, but to me that would just be a good result from a bad process: one that could just as easily go the other way (e.g. refusing blood from gay donors) in other circumstances. The solution here isn't to ram through *our* beliefs — it's to put the decisions back in the hands of the people who are most qualified to make them.

philipjohn:
It's important we're clear on what homeopathy is. From Wikipedia:

> Homeopathy ... is a form of alternative medicine created in 1796 by Samuel Hahnemann based on his doctrine of *like cures like* (similia similibus curentur), whereby a substance that causes the symptoms of a disease in healthy people will cure similar symptoms in sick people.

The use of the word "doctrine" here is important. It is true that treatments can work when derived from the cause of the disease. The difference between homeopathy and medicine is that medicine only uses this method as a treatment where it is proven, through the scientific method, to work whereas homeopathy is the *belief* that this method works, regardless of evidence.

For that reason I believe the PR is correct. If a "homeopathic" treatment is put through the rigour of medical science, and proven to work, it ceases to be homeopathy and becomes medicine.

So yes we should, across the NHS, remove unproven treatments based on the homeopathic doctrine and instead use public funds on treatments scientifically proven to work (that includes, by the way, placebos which can form a legitimate form of treatment).

tmtmtmtm:
@philipjohnm, if you can find a way to turn that into a general policy around requiring X amount of evidence before more than Y amount of public funds gets spent on something, then (other than perhaps some back-and-forth over the specific numbers), I doubt I'd I'd have much hesitation in voting for it. (Aside: does the UK even have a law requiring the publication of all clinical trial results yet? If not, that sounds like a much more useful thing to introduce.) But in the absence of a significant problem, it's **much much much** more important that doctors, not politicians, are responsible for medical care decisions. 

If we really want "something new", it should be to vigorously oppose politically-driven health-care decisions, not merely to impose a _different_ set of ideological ones.

philipjohn:
> does the UK even have a law requiring the publication of all clinical trial results yet

It doesn't! But it should.

> it's much much much more important that doctors, not politicians, are responsible for medical care decisions

I think we have that already in the 'NHS constitution' part of the manifesto.

We should perhaps have the following instead: we'll lobby for the NHS constitution to include the requirement that only scientifically proven medicine may be provided on the NHS, and that all trials must be public to allow medicines to be scrutinised effectively.

tmtmtmtm:
> I think we have that already in the 'NHS constitution' part of the manifesto.

I'm guessing this is this part? "Politicians cannot be trusted with the NHS. We will make the NHS independent, governed by a constitution created by a citizen-led constitutional convention."

Perhaps worth tightening the language of that first part a little more? Or is it deliberately covering a wider range of potential issues? I'd like to see something more explicit around the concept that it should be doctors making medical decisions, not politicians.

> We should perhaps have the following instead: we'll lobby for the NHS constitution to include the requirement that only scientifically proven medicine may be provided on the NHS, and that all trials must be public to allow medicines to be scrutinised effectively.

I'd be happy to support that.

philipjohn:
> I'd be happy to support that.

Nice :) 

What do you think to all this @andrewdwilliams ?

andrewdwilliams:
I was kind of sitting back and ignoring this, actually, @philipjohn, but I'll throw in my two cents.

## Two cents

[This article](http://www.theguardian.com/commentisfree/2015/mar/12/no-scientific-case-homeopathy-remedies-pharmacists-placebos), by the Professor for Complementary Medicine at the University of Exeter, sums up his findings through several trials, surveys and reviews of homeopathy:

* Our trials failed to show that homeopathy is more than a placebo.
* Our reviews demonstrated that the most reliable of the 230 or so trials of homeopathy ever published are also not positive.
* Studies with animals confirmed the results obtained on humans.
* Surveys and case reports suggested that homeopathy can be dangerous.
* The claims made by homeopaths to cure conditions like cancer, asthma or even Ebola were bogus.
* The promotion of homeopathy is not ethical.

Furthermore, what is possibly the "most thorough and independent evaluation of homeopathy in its 200-year-long history", [a study](http://www.nhmrc.gov.au/guidelines-publications/cam02) by the Australian National Health and Medical Research Council reached similar conclusions.

This is their [summary statement](http://www.nhmrc.gov.au/_files_nhmrc/publications/attachments/cam02_nhmrc_statement_homeopathy.pdf):

>Based on the assessment of the evidence of effectiveness of homeopathy, NHMRC concludes
that there are no health conditions for which there is reliable evidence that homeopathy
is effective.

>Homeopathy should not be used to treat health conditions that are chronic, serious, or could
become serious. People who choose homeopathy may put their health at risk if they reject or
delay treatments for which there is good evidence for safety and effectiveness. People who
are considering whether to use homeopathy should first get advice from a registered health
practitioner. Those who use homeopathy should tell their health practitioner and should keep
taking any prescribed treatments.

>The National Health and Medical Research Council expects that the Australian public will be
offered treatments and therapies based on the best available evidence.

You can read the other information provided by the NHMRC in your own time, all the information is there.

## The point

Taking aside what has been said above about evidence based policy, this was opened to discuss NHS spending on homeopathy and that alone. If we're to follow what's written in the [Introduction](http://openpolitics.org.uk/manifesto/index.html), and have a government that is "Rational, with decisions and policy based on sound evidence, not media pressure", then what we must do is obvious. There is no place for homeopathy in an evidence-based government, therefore there is no place for homeopathy in an evidence-based publicly-funded health service. It is that simple. 

In this instance, central government should override the wishes of front line staff, in the interests of aiding parents, because, as the NHMRC study found, "People who choose homeopathy may put their health at risk if they reject or delay treatments for which there is good evidence for safety and effectiveness." This is not ideological as @tmtmtmtm claims, this is science and evidence based, in the interest of patient safety above all else.

No one's quite sure how much the UK spends on homeopathy a year, although the British Homeopathic Association [puts the figure](http://www.britishhomeopathic.org/what-is-homeopathy/homeopathy-and-the-nhs/) at around £4 million. [This article](http://www.theguardian.com/society/2009/jun/10/complementary-medicine-nhs-more4) suggests closer to £12 million. As I said, no one's quite sure, but the figure is certainly in the millions, and as @Floppy mentioned there are also three homeopathic hospitals on the NHS. Is there a solution in privatising fully the homeopathic hospitals? Perhaps.

I hope, @tmtmtmtm, you realise why this demands a separate topic. Because this issue is of the utmost importance, with people's well being put at risk, and because people will be reassured by our stance of taking unequivocal action against homeopathy on the tax payer.

tmtmtmtm:
> I hope, @tmtmtmtm, you realise why this demands a separate topic. 

No, sorry, I don't. Even if the figure is £12 million, that's not even a rounding error in a budget of ~£115 billion. I'd have no problem with a proposal that infrastructure (like hospitals) be repurposed or privatised, but much as I personally disagree with homeopathy, I much more strongly disagree with the idea that this is a question for politicians, rather than doctors.

andrewdwilliams:
This is not really a question for politicians. This is a question for scientists and professionals in the field of Medicine, with a vast wealth of experience greater than anything you or I may have. The scientists have addressed the issue on many, many occasions, with the NHMRC's study being the most comprehensive of all. Their advice is clear - homeopathy does not work. It is bad, borderline dangerous. It doesn't take much to take their hint. Doctors have made this decision, not politicians.

This is why it should be included in the manifesto. On this occasion, I'm sorry to say, the evidence is stacked against NHS doctors and hospitals that practice homeopathy. They are wrong and other doctors are right. It's a matter of fact. The amount of money is of no consequence, for sure, but the matters of morality, ethics and basic principles are. That is why this issue has to be singled out and why we must include it in the manifesto. There is not a single justification for homeopathy.

[Doctors](http://www.theguardian.com/society/2010/jun/29/ban-homeopathy-from-nhs-doctors) [don't](http://www.theguardian.com/science/2010/oct/27/nhs-funding-homeopathy-chief-scientist) [want](http://www.dailymail.co.uk/health/article-2267120/Homeopathy-rubbish-shouldnt-available-NHS-says-chief-medical-officer.html) [homeopathy](http://www.bbc.co.uk/news/uk-scotland-11277990).

In this case, science, fact, safety and ethics, trump local decision making.

tmtmtmtm:
@andrewdwilliams Would you support a proposal to make it illegal for doctors to prescribe a placebo?

andrewdwilliams:
Perhaps, if only the on the grounds that it is unethical to prescribe medicines that are not what you say they are. I would most certainly outlaw placebos when you don't end up taking medicine that actually helps your condition, therefore endangering yourself. Studies have shown this is what happens with homeopathy in particular.

A better understanding of the placebo effect, including whether or not it is actually "regression to the mean" before a proper answer can be given as to whether placebos are worth the money and actually ethical.

tmtmtmtm:
And yet studies show that almost 100% of GPs prescribe placebos, presumably making this **much** more of a pressing issue than homeopathy?

However both issues are currently a matter for local Trusts and GPs, not central government, and the manifesto already promises to make the NHS completely independent, with a citizen led constitution. Promising to legislate on either of these issues is in direct contradiction of that. Approving this should only happen if that is removed.

andrewdwilliams:
Fair enough. Although we have no idea what form an NHS constitution, I think it would be fair to add a different suggestion to the manifesto instead of this one.

"We will conduct a review of placebo treatments in the NHS, such as homeopathy, to decide whether placebos are cost-effective, ethical and effective at treating patients."

andrewdwilliams:
@tmtmtmtm I've made the alterations. If you approve, give it a thumbs up, because you need to remove your abstention before it can be accepted. But I'm not sure if technologically this will work. @Floppy, advice?

Floppy:
:+1:

Floppy:
@tmtmtmtm's abstention won't block the merge, just make it need one more vote than normal to override. However, hopefully he will now agree with the content and can upvote instead :)

tmtmtmtm:
> hopefully he will now agree with the content and can upvote instead

No, I'm afraid I don't. Much as I agree with the hoped-for outcome of this, it still goes against the much more more core principle of setting the NHS free from a constant barrage of politically motivated policies. If we're saying that it should only be free from the sort of interference we happen to disagree with, then we should say so explicitly instead of trying to pretend that we're any different.

Floppy:
@tmtmtmtm I do see where you're coming from on that. It's tricky. I'm mulling over something that might capture the combination of a desire to only spend money on evidence-based treatment, while leaving the NHS free to make its own decisions. Hmm.

andrewdwilliams:
> No, I'm afraid I don't. Much as I agree with the hoped-for outcome of this, it still goes against the much more more core principle of setting the NHS free from a constant barrage of politically motivated policies. If we're saying that it should only be free from the sort of interference we happen to disagree with, then we should say so explicitly instead of trying to pretend that we're any different.

Whilst this was originally about homeopathy, I would like to make the point that it is absurd to think we can separate the NHS from politics when 18% of our government budget is spent on it annually. If we remove the influence of politicians, e.g. MPs, on the NHS, then we remove any sort of democratic oversight the public have on 18% of where the government budget goes. And, although this goes against the current trend in recent British political history, sometime politicians can have good ideas about how the NHS is run and how they spend some of their money. And whilst we may remove politics from the NHS, this does not automatically mean the doctors make the decisions, it will still be very much top-down, and run instead by senior civil servants, some of whom may well be even further detached from reality than politicians are.

I would also suggest that having a review into placebo treatments is hardly political interference - for a start, I did not even mention whether or not we would act on the results of the supposed review. As I've already stated, this would also not be an ideology-based review of any kind, unless backing up policy decisions with evidence is an ideological thing to do. I fundamentally disagree that the policy I have suggested is "politically motivated", if only that I think we should explore whether there are more responsible ways to spend our money. That is not for me to decide, however, which is why I suggested a review.

philipjohn:
We're conflating two things here;

* Government setting the principles under which the NHS should operate
* Government interfering in the day to day running of the NHS

Setting the principle that the NHS shouldn't waste money on unproven medicine is an example of the first of those two.

DoH setting junior doctor's contracts is an example of the second.

So, yes, we should absolutely set the principles under which the NHS operates and that should include something as obviously valid and important as rejecting unproven medicine.

tmtmtmtm:
That seems back to front to me. The basic principle is that medical decisions should be made by doctors, not politicians.

philipjohn:
It's not a medical decision - it's a safeguarding decision. Doctors still decide the medicines they prescribe but they do so *from a pool of allowed medicines already*. We're merely saying that pool should not include unproven medicine.

tmtmtmtm:
And that pool of allowed medicines should be set by the medical community — not by politicians.

philipjohn:
Actually, what we're talking about is done by NICE, which falls under the Department of Health, and already looks at evidence relating to homeopathy. In reality, the result of this policy would be to ensure that clinicians are actually following the NICE guidelines based on their assessment of the evidence.

Floppy:
This is a great conversation, and I want to make sure we get to a result one way or another. Can anyone propose an alternative formulation of the ideas here that handle the independence of the NHS while adding the view that we only wish to spend money on medical treatments that are scientifically proven? Perhaps the NHS could provide alternative treatments if it wishes but they would be treated as "elective", so paid for by the patient? Feel free to open another PR if you want - we'll link it into the main conversation here.

Floppy:
@tmtmtmtm I certainly agree that the treatments available should be set by medical professionals not politicians, by the way.

philipjohn:
Of note: https://richarddawkins.net/2015/11/homeopathy-could-be-blacklisted/

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/363), and results are set in the merge status checks below.

## Changes

PR #366, opened 2015-10-03
Transport: allocate road space to bus lanes

h2g2bob:
My suggestion is to encourage frequent service where there is appropriately dense land use ("towns and cities") as this will reduce traffic and congestion. "Allocating road space" will mean taking space away from other traffic. Making bus journeys faster improves both travel time and time waiting for the next bus (it's a double benefit).

Congested roads exist because there's not enough space for all the vehicles (this is true, even with electric cars). I'd like to reward people who are reducing traffic by using public transport.

I think this policy is consistent with other policies in this section, which encourage a shift to less-polluting modes of transport.

I'm influenced by [Human Transit](http://www.humantransit.org/2015/07/mega-explainer-the-ridership-recipe.html) in this suggestion.


andrewdwilliams:
:+1:

PaulJRobinson:
:+1:

Floppy:
:+1: and we should extend to cycling in a followup as well.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/366), and results are set in the merge status checks below.

## Changes

PR #367, opened 2015-10-30
Introduction for economy section

andrewdwilliams:
Introduced a brief summary of our actions to the economy section. Subject to significant change. More of a placeholder for a refined statement.

### DIFF
--- a/manifesto/economy.md
+++ b/manifesto/economy.md
@@ -5,7 +5,7 @@ title: Economy
 * table of contents
 {:toc}
 
-We think the economy should work for everyone, not just billionaires, hedge funds and investment banks. We oppose the austerity policies of George Osborne that have doubled the national debt in the last five years and caused much unnecessary suffering for those on the lowest rungs of the economic ladder. We want to stop tax evasion and avoidance, so that everyone pays their fair share. We want to introduce a Land Value Tax to replace Council Tax, Uniform Business Rates and Stamp Duty. We want to introduce a Financial Transaction Tax, and we also want rent controls. We want to raise and minimum wage, and we encourage everyone to pay the living wage, starting with the government.
+We think the economy should work for everyone, not just billionaires, hedge funds and investment banks. We want to improve fairness in the tax system by closing tax evasion loopholes and reducing the gap between corporation tax and income tax. We want to create affordable housing for everyone, and to do this we will introduce a land value tax, adjust capital gain tax and introduce rent control. We want to help the lowest paid and we will by raising the minimum wage and encouraging everyone to pay the living wage.
 
 ## Minimum Wage
 
--- a/manifesto/economy.md
+++ b/manifesto/economy.md
@@ -1,13 +1,11 @@
 ---
 title: Economy
-layout: policy
-published: true
 ---
 
 * table of contents
 {:toc}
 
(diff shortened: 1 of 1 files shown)


tmtmtmtm:
> We oppose the austerity policies of George Osborne that have doubled the national debt in the last five years

Do you have a reference for that?

http://www.ukpublicspending.co.uk/spending_chart_2005_2015UKr_15c1li111lcn_G0t_Total_National_Debt#view has national debt rising from ~£0.96tn to ~£1.46tn over the last five years, which is about a 50% rise, rather than a doubling. (update: FullFact have also [covered this one](https://fullfact.org/factcheck/economy/has_uk_debt_doubled-41769))

Framing this as being a result of austerity is also rather odd, as the other side of that argument is that austerity is what has stopped the debt from being much larger than it actually is. If we want to claim that it actually *has* caused it, then we need to be much more explicit as to how, otherwise we need to be clear whether the position is that (a) we _should_ have been OK with growing the debt, because the results of austerity have been much worse than the results of extra debt would have been, or (b) we should have tried to reduce the debt in a different way.

andrewdwilliams:
How about "We oppose the austerity policies of George Osborne and believe that his response of cut, cut, cut was the opposite of what was needed following the financial crash. Instead, we believe that a sensible program of economies combined with large-scale investment in infrastructure and house building would have created growth quicker, resulting in less job losses and less cuts to the welfare budget." 

I'm just floating ideas here. We don't really have a central theme to our economic outlook so I'm trying to create one. I assume that we want to adopt a broadly Keynesian economic outlook, as in "invest rather than cut", but I'm conscious that we're now at a stage where the 2008 crash is less and less relevant and now we ought to be looking at how to introduce growth and how to cut the deficit in a sensible manner.

h2g2bob:
"Remove Tampon Tax" can probably be included under the "VAT" section (which is the tax involved)

h2g2bob:
> We think the economy should work for everyone, not just billionaires, hedge funds and investment banks.
I like this, it's a statement of intent.

> austerity
I approve of this, mostly, but it goes further than just re-wording the introduction. If we adopt this, it should certainly be a full section on the page and not just a comment in the introduction. It should probably be its own merge request.

Except for the "austerity" part, the rest of the introduction is a fair summary. But it doesn't state the intention behind the policies.

I see the core of this page being:
- improving fairness in the tax system by closing tax evasion loopholes and reducing the gap between corporation tax and income tax
- creating affordable housing by introducing a land value tax, adjusting capital gains tax and introducing rent control
- helping the lowest paid by raising the minimum wage

I would love to see the introduction wording changed along those lines.

andrewdwilliams:
@h2g2bob I have made the alterations. Let me know what you think.

Floppy:
:+1: I like this, except we want to make the minimum wage the same as the living wage, not just encourage it, so we could be stronger and say "make sure every citizen can earn a living wage" or something.

I might propose some wording changes later, but I'm OK with the general thrust of it.

h2g2bob:
:+1: Looks great

andrewdwilliams:
@floppy Could this be merged?

Floppy:
Yep, can now. Will do so.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/367), and results are set in the merge status checks below.

## Changes

PR #374, opened 2015-11-23
Prime Minister line of succession

andrewdwilliams:


Floppy:
Another horrible editor clash is breaking the diff here. The relevant text for reviewers is:

## Line of succession

Introduce legislation setting out the line of succession for the Prime Minister in the unlikely but
possible event that they are incapacitated and unable to govern, in a similar style to that of the
Presidential Line of Succession in the United States. We could, for instance, introduce a variation
on the [Prime Minister (Replacement) Bill](http://www.publications.parliament.uk/pa/bills/cbill/2013-2014/0059/14059.pdf).

Floppy:
So, this seems sensible, and something you would have in a constitution (the core OS of the country), perhaps. My question is, do we not have this at the moment?

andrewdwilliams:
No, we don't. Our constitution is very much a make-it-up-as-you-go-along arrangement, which would be precisely what we would do in this situation.

I was just thinking, though, if this would be better going under the constitutional convention section, as something like: "during the convention we would consider the introduction of a codified line of succession for the PM in case they are incapacitated."

Floppy:
I'm a :+1: on adding this, btw. Sorry, forgot to vote.

Floppy:
Oh, sorry, I renamed the branch and it auto-closed this request. Can you resubmit it? Sorry about that.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/374), and results are set in the merge status checks below.

## Changes

PR #379, opened 2016-05-10
This change is to bring into the discussion the latest 'discoveries' regarding natural learning.

Corneilius:
Learning by exploration and critical analysis is a fundamental biological mandate. It is how infants learn to walk, use their arms and hands, talk and many other key skills. The fact that compulsion schooling exists is due to a crazy distrust of children's motivations and abilities, a distrust that predicts itself, that is a self fulfilling prophecy, the origins of which date back to ancient pre-Classical Greece - 'Spare the Rod, Spoil the Child'.

If we are to build a healthier future, then we have to start with the way we relate to infants, children, adolescents and parenting.

The psychology of the system we have today is the result of how the adults that operate it were treated as infants, children, adolsecents and young adults.

To put it another way:

"The psychology of any given family, community or society is both revealed and perpetuated in how that family, community or society relates to and treats the children and other vulnerable people."

### DIFF
--- a/education.md
+++ b/education.md
@@ -51,7 +51,9 @@ We recognise a wide variety of psychologies and perspectives amongst our childre
 
 ## Separation from Militarisation.
 
-We suggest that the Military Cadet system be removed from all schools, and that History Education be 100% honest and transparent in all respects. We sense that would go a long way to combating the emergence of right wing support across the country.
+We suggest that the Military Cadet system be removed from all schools, and be replace with something that provides the opportunity to learn and explore the same skills divorced from Militarisation. Military activity is about killing, and has no place in a rational, humane environment. Enlistment in the armed forces ought to be something chosen by adults who are well informed, mature and resilient, rather than children or teenagers who can have only a fantasy perception of the realities of military action.
+
+As part of that we also suggest that History be an exploration, rather than rote learning of 'approved' texts. Children ought to be allowed to explore all available materials as individuals, then encouraged to bring what they have found back to share with their fellow students, and to discuss what they have found, in order to arrive at a more nuanced understanding and to help the children to develope their innate critical thinking skills in a group setting. Varied perspectives can then be integrated. We would look to the quality of their logic and research, their ability to define, integrate and refine, in written and spoken word, as the measure of their progress.
 
 [^1]: [British Social Attitudes Survey](http://ir2.flife.de/data/natcen-social-research/igb_html/pdf/chapters/BSA28_12Religion.pdf) (pdf)
 [^2]: [Maintained Faith Schools, Department for Education](https://www.gov.uk/government/publications/maintained-faith-schools/maintained-faith-schools)
--- a/education.md
+++ b/education.md
@@ -41,7 +41,21 @@ All state secondary schools will have an elected students council, and student p
 
 We will remove the guidelines[^4] that force teachers to act as counter-terrorism officers and that stifle free speech within the learning environment. 
 
+## Evidence based Learning
+
+We recognise the tension between the coercion inherent in 'being taught' to follow a curriculum and testing regime, and the innate drive of self motivated learning, which is a biological mandate, where the student requests specific help to explore a given subject area, based on their interest and resources. [^5]
+
+We therefore suggest that all teaching staff, support staff in Education be trained in the ethics of Education without compulsion, and furthermore, that all teaching staff undergo training in empathy, conflict resolution, open democracy, child development and other areas aligned to the welfare and healthy development  of our children as people, as the future adults of this Society. How we treat the child determines the emerging psychology of the future.
+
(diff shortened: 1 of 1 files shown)


Floppy:
I am not an educator, so I don't like to express solid opinions on it, but as a parent I like the approach here. 👍
I suspect the 100% accurate history thing is, while a nice idea (and again I agree with the sentiment), unrealistic. Is there even such a thing as 100% accurate history?

andrewdwilliams:
I agree with the three paragraphs under 'evidence based learning', but cannot support the other paragraph. Firstly, I have gone through the 'military cadet system' in my school, ie the CCF. I thought it was great, probably one of the best things I've ever done through school, and it taught me a lot, including stuff I don't know it taught me. I enjoyed it immensely, and would not support removal. I also don't see a correlation between the CCF and right wing thinking.

I also agree with @Floppy that 100% honest history does not exist - to an extent, that is why we study history. 'Honest' history simply does not exist - why else would so many theories exist, or different viewpoints about the causes of this, or the consequences of that. It is not an exact science and so cannot be pinpointed down to certain facts or theories that someone perceives as the 'honest' account of events. If the second paragraph was removed, I could support the otherwise sensible proposal.

OliverJackson1:
I've always thought that there should be an assessment every year on the best strategies to teach children. As a variety of learning styles exist it seems crazy not to allow this to develop. I was always a kinetic learner growing up and loved learning through games. A friend of mine though developed very well through Auditory means (my most hated - can't sit still). So all in favor of the Evidence based learning so long as it allows those that don't respond to that to be equally accompanied.

In terms of honest history, I don't like this. You run a terrible risk seeking to clean history of opinion, who arbitrates, who delivers, who enforces; very messy. I also, as with @andrewdwilliams don't think its possible. 

I believe that everything that is a lifestyle choice being removed from educational platform, forming only aspects of the curriculum. As such, to me, militarism is a lifestyle choice. Therefore, in agreement with you @Corneilius  on that. That said, despite there being an arguable association between military participation/support and right-wing politics, I don't think that its endearing or rational to put it in a manifesto. 

So, I'd drop the "Honest History" and reference to the right wing. For me cadets remain but turns to become a decision by the parent out of school time, we aren't going to stop people to choose to follow what they want to, are we?

Corneilius:
With regard to an 'honest history' three threads occur to me.

1. Telling the story from the point of view the official authorities who hold power.
2. Telling the story from the point of view ordinary folk who lack power.
3. Telling the story in human terms, in terms of real outcomes in peoples lives. 

example : The Battle of The Somme

1. British Official History of The Battle of the Somme, German Official History of The Battle of the Somme, and others who were involved at the High Command and Political Level, their justifications for what they initiated and ordered, their propaganda etc... and their 'rewards' or 'losses'.

2. The story of the young men who either enlisted or were conscripted. Their experience and outcomes. How they responded to the trauma, how that affected their home lives and relationships.

3. The story of civilian populations displaced as the war waged this way and that way. 

The idea is to enable children to explore the available data and come to their own conclusions, which they would then share, and through discussion reach a wider understanding. 

We know that a natural quality in autonomous learning occurs when children seek out information and share it amongst themselves, as a co-operative venture. What one learns,the entire group learns. This dynamic is quite efficient. So rather than rote learning of official texts, the children seek out information, check it, bring it back, check it against what others have found and discuss the matter. This dynamic also tends to reduce peer pressure, as differences are appreciated.

I fully realise that 'opinions' exist and are often hard held positions, often the result of a self limiting range of materials published by Authorities with an agenda.

Corneilius:
I accept the comments on 'right wing' and 'militarism', and I believe that my clarification would tend to reduce polarised positioning. Which is the point.

As to Cadets, the key aspects for children are the physicality, the team work, the bonding and the skills developed - surely we can provide a similar environment without the militarized aspects.

OliverJackson1:
I appreciate the update and the clarification that it provides. Please correct me if I'm wrong but your point about the cadets is that it is inappropriate for children as they are by all accounts vulnerable to the influences of external forces. 

If I've not become confused here, I would advise amending the text to focus on the development of the child outside of external lifestyle influence so they can be independent. I understand and empathise with your wording currently but some will find that divisive as the military is part of their identity. We don't want to alienate people and "Military activity is about killing, and has no place in a rational, humane environment." is quite the statement! One that even with empathy I can't agree with 100%. The military and militarisation have purpose outside of killing.

I like the entry for history but think this would be applicable to nearly all non-science subjects, its effectively independent learning. 

I also do not see the relevance to separation from militarisation. With that in mind, could this section be 'Independent Learning'? To remove external influence from the education environment and to promote the independent development of children?

philipjohn:
This proposal has exceeded the time limit, so we'll close it off, but @Corneilius please do think about what could be done to make this proposal more successful, and try again. Thanks!

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/379), and results are set in the merge status checks below.

## Changes

PR #381, opened 2016-06-27
Stopping unelected ministers

ranrobinson:


Floppy:
👍

Floppy:
Agreed 👍

tmtmtmtm:
What's the rationale for this?

Floppy:
It's a tricky question, but one of the major components of the EU referendum was a pushback against "unelected" people in the EU commission and council. 

While not exactly the same, a similar argument could be made for not giving government posts to unelected people here. Even though we don't elect our government directly anyway, but that's a bigger issue that I think we cover elsewhere.

andrewdwilliams:
I appreciate the sentiment behind this, but in my view the idea is far too restrictive. Often there are complaints about the quality of government ministers, so I would dread to think how far that would drop without peers to support them. This could have a serious effect on the working of government. People like Lord Falconer, Baroness Warsi, Baroness Anelay, etc., all hold important roles in government.

The House of Lords is not a house of experts - I accept that, but some of them are experts, and beyond that, many of them are very experienced in the workings of government and legislation, more so than recently-elected MPs. I do not support the unelected House of Lords, but barring them from becoming ministers until we abolish/reform the House of Lords is not helpful.

Instead, I believe we should strive towards a system used in many European countries, one, for instance, being Greece. That is a system where both MPs and people who are not elected can be appointed as ministers. However, even those who are not MPs are accountable to Parliament, and have to stand up and give speeches, etc. Incorporating the system used in the US, non-MP cabinet ministers could be approved by Parliament before assuming their role.

This not only makes it even easier for experts (such as university professors - there are many in the Greek cabinet) to join the government, but it increases the separation of powers between the legislature and the executive, which in my opinion is a good thing and is something we should strive towards.

philipjohn:
:+1:

Floppy:
I'm going to change to ✋ for now while I think on the points raised; I agree that we should be able to appoint the best people to the executive, and it should be separate from the legislature. Going to think about whether this is a short term interim solution or not.

tmtmtmtm:
I'm far from being a constitutional expert, and the precise set-up of the relationship between Parliament, the Cabinet, and the Privy Council, is something I'm not entirely clear on¹, but in general I'm in favour of a stronger separation of powers, so I get a little nervous about what seems to effectively be a proposal for Parliament to get _more_ involved in how the Cabinet is formed (i.e. this is presumably saying not just "If we formed a Government, we would not appoint any ministers who aren't members of the House of Commons", but the much stronger, "We would pass legislation to prevent future Governments appointing ministers who aren't members of the House of Commons")

(And although there is no requirement that the Leader of the House of Lords _must_ be a member of that  house, it would seem very odd to say that they must _not_ be!)

---

¹ Though pointers very welcome, as it's something I'd definitely like to understand better.

philipjohn:
It's entirely possible that a Prime Minister could appoint a bunch of new Lords and then form a Government composed entirely of unelected Lords. In fact, in reading around on this issue I've discovered that Ministers don't technically have to be members of *either* house, which is astonishing.

This, for me, would be an interim thing, and I think it reads that way with the "Until the House of Lords is abolished" caveat.

People will have much more faith in government knowing that every single member of it is elected, IMO.

andrewdwilliams:
@philipjohn Thus is the nature of the UK's constitution - if it hasn't specifically been ruled out, it's possible.

See my earlier comments for why I disagree with this being an interim arrangement - it would mean a potential loss of actual talent, and it would make for worse governance (in the interim) as government would be comprised solely of MPs, including those who are serial rebels or plain incompetent.

I don't think being elected will be the thing that instills confidence in government. I think being competent would be the thing that instills confidence in the government. Competence is not arising in the present arrangement, but it would be even worse under the interim arrangements you propose.

philipjohn:
This proposal has exceeded the time limit, so we'll close it off, but @ranrobinson please do think about what could be done to make this proposal more successful, and try again. Thanks!

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/381), and results are set in the merge status checks below.

## Changes

PR #384, opened 2016-06-28
Added a new section

OliverJackson1:
Media retractions are rare. Mostly an outlet will just retrospectively delete the article from their back catalog. This isn't acceptable as it subverts the public dialogue. This website is good for tracking the Sun alone but all outlets are guilty of this http://the-sun-lies.blogspot.co.uk/2009/01/echo-chamber.html. 

It was asked that the request stay small so I've shy'd away from a further punishment for a number of consecutive retractions but I would like to introduce this too you at a later point. I would suggest 3 retractions in 3 months requires a fine and a headline, dictated by a regulator of the retractions, this would take the prominent position in the News Outlet.

I would also suggest that we consider enforcing fines for article deletions, instead of retractions. 

Be good journalists, if you fail, retract the article. If you fail frequently, its by choice or you need a new profession, either way, stop it! :)

### DIFF
--- a/culture.md
+++ b/culture.md
@@ -17,3 +17,7 @@ Immediately reduce the term of copyright from '70 years from the end of the cale
 Encourage the Advertising Standards Authority to tighten regulations around the use of pseudo-scientific language and terminology for the promotion of cosmetic, toiletry, food or other products, without sufficient evidence.
 
 [^1]: [Can we have our freedom to joke now please?](https://www.openrightsgroup.org/campaigns/modernise-copyright) - Open Rights Group
+
+## Media Retractions
+
+Media outlet retractions should mirror the original article. This covers font size, word count, page space and positioning within the document. If a media outlet does not follow this it will receive a fine from OfCom.
--- a/culture.md
+++ b/culture.md
@@ -20,4 +20,4 @@ Encourage the Advertising Standards Authority to tighten regulations around the 
 
 ## Media Retractions
 
-Media outlet retractions should mirror the original article. This covers font size, word count, page space and positioning within the document. If a media outlet does not follow this it will receive a fine from OfCom.
+Media outlet retractions should mirror the original article. This covers font size, word count, page space and positioning within the document. If a media outlet does not follow this it will receive a fine from Independent Press Standards Organisation (IPSO).


Floppy:
👍

glynnchaldecott:
Alternatively a system could be implemented whereby all news agents need to submit all articles to a central government site, within a specified period of publishing for permanent storage. The site would be cataloged and searchable by the general public. This would effectively eliminate the ability of news agencies to secretively retract articles, and safeguard against agencies closing resulting in their sites being switched off (rare for the big boys I know but not unheard of.)

Floppy:
@glynnchaldecott I reckon you could do something like with a long-term archive run by the press regulator, perhaps in conjunction with the Internet Archive, possibly even in a fully decentralised way. There are some people looking at things like this to be able to provide permenant links to press content. Keeps it away from central government a bit too :)

But, on this specific aspect of corrections in print media, I'm in agreement. :+1:

andrewdwilliams:
@OliverJackson1 I support this in principle, and it can of course be added to and refined at a later date, but one small point of order before I upvote this. OfCom is the regulator of TV and radio, not newspapers. The Independent Press Standards Organisation (IPSO) is the main newspaper regulator, but membership of it by newspapers is voluntary, which touches on another issue. I don't know whether IPSO currently has the statutory power to introduce fines either. Essentially, as a minor issue I support this, but the broader issue of press regulation also ought to be addressed somewhere in the manifesto.

philipjohn:
:+1: 

I have a feeling that the British Library does archive newspaper content. Not so sure about online stuff, although I know that the very very small and local news site I help run was approached about being archived by them.

andrewdwilliams:
:hand: Just on a point of order, I'm going to block this until we get OfCom changed to IPSO (or unless James can alter it post-acceptance). It's a small point, but otherwise makes us look like we have no idea what we're talking about.

Floppy:
Agreed, I will make that change after merge, unless @OliverJackson1 wants to do it first. 

Oliver, if you want to, the "in progress" editing is a bit of a pain, but you can do it by hitting the pencil icon on https://github.com/openpolitics/manifesto/pull/384/files.

andrewdwilliams:
I'm happy to change to a :+1: if James will edit that in. That was all that really concerned me.

Floppy:
👍

OliverJackson1:
Lots of comments! :). @glynnchaldecott I don't mind the current method of citizenry watchdogs. The new system you write about would require bureaucracy and to my mind we should avoid bureaucracy wherever possible. @andrewdwilliams Valid point. My apologies for submitting with the wrong regulatory body. From my understanding IPSO can fine to a maximum of 1% of turnover. This hasn't happened as far as I'm aware. In line with your request I have updated the wording. Once this is accepted, shall I draft a more developed Press Regulation section? Then we can amend as our discourse determines?

Floppy:
@OliverJackson1 any further additions you'd like to make once this goes in would be awesome :)

andrewdwilliams:
I think I have to thumb it up again. :+1:

tmtmtmtm:
Other than for the simplest of corrections, this seems guaranteed to ensure that media outlets simply refuse to retract anything that they're not forced to (thus having the perverse result of lowering rather than raising overall journalistic standards). And if we're talking about *forced* retractions then we need to get a lot deeper into the mechanisms for that. At the minute this section reads as if Leveson (and the surrounding debate about press freedom) had never happened.

philipjohn:
> At the minute this section reads as if Leveson (and the surrounding debate about press freedom) had never happened.

I'm struggling to see where that comes from. This was actually in the Leveson report.

:+1:

tmtmtmtm:
> This was actually in the Leveson report

If all that is being asked for were what was in Leveson, then this would be unnecessary, as IPSO already _has_ the power to require prominent corrections and apologies. 

But the proposal here goes much further than that, to an extent that I think it would be counterproductive. This change needs to make a case for why the current situation is insufficient, and why this would be better. In the absence of that, this seems like an illiberal overreach to me. Governments should always err on the side of less interference with the press, not more. Like all freedoms, press freedom can certainly be abused — but we need to be very careful not to overreact when it is. The proposal here goes much too far IMO, so it's a 👎  from me for now.

OliverJackson1:
@Floppy I've agreed to the CLA but its still showing as if I haven't.

@tmtmtmtm the Leveson Report concluded: "a cultural tendency within parts of the press vigorously to resist or dismiss complainants almost as a matter of course"; this is the current situation. We don't exactly have editors running around issuing retractions. Retractions when they occur are frequently buried away from general sight. Worse even, many outlets simply delete the old content. This motion seeks to establish a third party remit for enforceable and consistent behaviors. Behaviors that do not currently exist.

There are examples everyday of this failing but currently the media portrayal of the leadership of Labour leader Jeremy Corbyn is most pertinent. Tom Watson was stated as having called for JC to stand down as party leader. This ran until Watson contacted the Beeb to state this wasn't true. The article was then amended. The article should have been met with a full retraction as the principle here is that facts and non-facts are daily being espoused with little consideration for the difference.

The motion does not threaten the freedom on the press. To assert that is absurd. It simply holds the press to account.

Leveson also concluded that there is a "casual attitude to privacy and the lip service it paid to consent demonstrated a far more general loss of direction". We all feel this loss of direction when we read or watch the media. This motion, in conjunction with future motions, aims to re-establish direction into one of the most corrupted and dangerous parts of the British State.

Clearly there needs to be a motion in the manifesto that addresses retractions so to block this seems harsh.

tmtmtmtm:
> Clearly there needs to be a motion in the manifesto that addresses retractions so to block this seems harsh.

This seems to be falling into the "Something must be done. This is something. Therefore this must be done" trap.

I don't think anyone (here, or indeed anywhere) is denying that there is a problem. What's at issue is what the solution should be. And the proposal here seems to both largely ignore the current landscape (if it's taking as its starting point that the post-Leveson changes aren't working or haven't gone far enough, it should, IMO, say so explicitly, rather than just being a generic statement that could just as easily have been written 5 years ago) _and_ also goes too far for me. Nor have you addressed my suggestion that this approach would simply make the press even more unlikely to ever retract or correct anything.

OliverJackson1:
@tmtmtmtm I don't think that it ignores the current landscape at all but it feels that this is not going to be resolved and nothing constructive will come, so I'll look to withdraw.

Personally, I don't feel that this was going far enough, I'd like to see motions that address the capitalisation of the media. That secures a greater degree of competition among media outlets and secures that the majority of income comes from membership payments and not advertisements. I'd also like to see greater security for journalists, as well as freedom. 

Fines should be given for article deletion.

The principle missed here is the difference between press freedom and unrestrained self-governance. This sector needs governance across a wide spectrum of areas, is deeply corrupt and is doing untold harm to the population. The lack of governance is fundamentally undermining the freedom it should have. 

The motion doesn't move to harm liberty, seeks to set out established practice and offers pejorative punishment when not followed. 

It terms of the retractions becoming less frequent that will occur either when the editors feel they can defend the original article or will not receive punishment for the original article. This motion does not affect these existing points.

philipjohn:
@tmtmtmtm Using the down vote to block proposals should be reserved for proposals that are fundamentally opposed to the values of the Manifesto project. As this doesn't fit that test it seems unfair to block it in this way. If a majority of contributors support the proposal it should be allowed to go through. Will you consider changing your vote to `:hand:` to allow a proper consideration of this proposal, please?

tmtmtmtm:
Ah, yes, I keep forgetting the rather idiosyncratic setup here where `:hand:` means -1, and `:-1:` means stop!

✋

Floppy:
@tmtmtmtm now you put it like that, it is rather daft yes :)

philipjohn:
I think votebot might be wrong about this one needing more votes?

Floppy:
Yeah, it's miscounted your vote for some reason as a hold. How strange. The balance of votes is enough to get it accepted. Merging!

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/384), and results are set in the merge status checks below.

## Changes

PR #387, opened 2016-06-29
Overseas Territories section update

andrewdwilliams:
Updated some stuff, added a little more, just general expansion.

### DIFF
--- a/foreign_policy.md
+++ b/foreign_policy.md
@@ -59,24 +59,36 @@ We should use our significant financial position within the global economy to he
 
 ## British Overseas Territories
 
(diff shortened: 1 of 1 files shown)


philipjohn:
:+1:

OliverJackson1:
@andrewdwilliams are we comfortable enough with the actions of President Erdoğan to agree to simple de-listing of the PKK? Would we potentially be better calling for a Kurdish state as a form of peace agreements out of concessions from Iraq and Syria. Not sure that we should really mention the PKK at all just forward devloution and non-intervention. Something like "We wish to pursue an evenhanded foreign policy and in light of violence from both the Turkish state and the Kurdish Rebels we condemn both sides. We believe in the establishment of a Kurdish State in the North East and North West Regions of Syria and Iraq, respectively, where there are effective military campaigns against Daesh (ISIS). We believe this will establish peace and we aim to see this new state provisioned with peacekeepers until such a time as peace is secured."

The fear is that otherwise we validate the violence of one side and not the other. Additionally, this conflict will run for decades without an evenhanded approach that recognizes the Kurdish desire for a free state.

Floppy:
@OliverJackson1 the only thing that changed in the PKK section is the footnote numbering, not the policy, so if we want to discuss it, let's do it under a separate change before we get offtopic here :)

andrewdwilliams:
@OliverJackson1 That wasn't my original addition to the manifesto, as @Floppy pointed out the only bit I altered in the PKK section was the footnote numbering. But if you want to change the PKK section, feel free to start a new pull request and I'll give it a look over.

andrewdwilliams:
@Floppy @OliverJackson1 Could you look at this and give it a thumbs up or not, so we can get this passed?

Floppy:
👍

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/387), and results are set in the merge status checks below.

## Changes

PR #389, opened 2016-06-29
Provision for calling constitutional convention

andrewdwilliams:
A half-baked idea, but one I didn't want to have kicking round in my head. I'm not sure whether a constitutional convention should be how the constitution is first made, instead of a citizens' panel (not the same thing), or whether it should only exist for the purpose of making amendments.

### DIFF
--- a/democracy.md
+++ b/democracy.md
@@ -13,6 +13,8 @@ The British constitution should be collected into a single written document that
 
 The constitution could either be created by a citizens' panel, with approval and adoption subject to a referendum, or created as a "blank slate", with every addition done as an amendment by Parliament with a super-majority vote (e.g. 75% in favour).
 
+Provision for calling a national constitutional convention to alter the constitution should be made in statute law. The law should call a constitutional convention if a super-majority (75%) of members of the Scottish Parliament, the Welsh Assembly, the Northern Ireland Assembly, and the English Grand Committee (if no English Parliament exists) vote for such a convention to be held. In the case one is called, members should be elected on a non-partisan ticket under a proportional electoral system, and it should have free power to make amendments to the constitution.
+
 ## Party Funding
 
 Limit all donations and loans to political parties to a maximum of £5,000 from individuals who appear on the electoral register.


Floppy:
Could be both - one idea in the manifesto at the moment is that we start with a blank-slate constitution, and make everything an amendment. :)

andrewdwilliams:
I spotted that. It's an interesting one. I would imagine constitutional conventions would be rarities only called when there are a number of things to add or change. So maybe the constitution should be a blank slate but we should then add things first time round with a constitutional convention. And I definitely think that regardless, provision should exist for a constitutional convention anyway, as we don't know what constitutional crises may arise in the future.

EdCaplen:
I like the idea of a constitution. The convention, much like the American ones should very much be ran based off of mutual compromise to ensure that all are represented.

OliverJackson1:
I've got an idea on this, why not work with the South African approach and ask every citizen directly what they need, starting from blank slate (country name even? Would probably lead to Britty McBritface, so perhaps not - joke does not work with the word country.). 

We would implement a provisional constitution of our own design but only for so long as consultation occurs with all citizens. For the South Africans this was two years. For a country the size of the UK it would probably be one rotation of parliament. 

A constitution needs to be the direct product of the people to create the feeling of citizenship: all citizens should have there say, both in its creation, ratification (ref: 50:50) and amendment (ref: 60:40). As democracy protects the minority from the power of the majority, asking all directly is the only way of securing that every minority is covered, the smallest minority, is the minority of 1.

philipjohn:
I like the idea of calling the constitutional convention, though I'd prefer that it were a citizen panel. We might say "non-partisan" but the reality is that it'll be political/celebrity figures with the funds to launch a campaign that will be elected and I don't think self-selecting people who are likely to be unrepresentative of the population should be deciding on constitutional amendments.

Floppy:
There are things here I'd change (how the convention members are chosen mainly), but it's a useful addition to the constitution proposal, so I'm going to upvote and then submit changes to a sortition-based system later on. 👍

philipjohn:
That's very pragmatic @floppy :) I'll agree with that! :+1:

andrewdwilliams:
@Floppy Can this be merged?

Floppy:
It can! Thanks for the reminder. I really should make this automatic soon.

Floppy:
pinging the votebot, which hasn't removed this. Please ignore this comment! :)

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/389), and results are set in the merge status checks below.

## Changes

PR #390, opened 2016-06-30
Amendment to Government Contracts

OliverJackson1:
I might be a little strong on the Treason part but it is what it is. Its a betrayal of the citizenry, the citizenry are the state, its a betrayal of the state, its treason. Other than that, I think the general premise here is sound, if any one government is going to run our assets or win our contracts, we are. We are not here as a nation to supplement other countries costs of living. Although, I have always considered it a means of establishing self-sufficiency in the NHS - outsource it to private healthcare states and charge a nominal profit, utilizing the monopolies of scale to create revenue, undercutting rivals and establishing market dominance. Digression aside, I hope the motion stands well.

Evidence of a state owned "private firm": https://www.edf.fr/en/the-edf-group/dedicated-sections/finance/financial-information/the-edf-share/shareholding-structure

### DIFF
--- a/economy.md
+++ b/economy.md
@@ -71,6 +71,12 @@ Ensure worker-owned cooperatives and mutuals are given priority with all governm
 
 All private companies who receive public money must be subject to the same transparency requirements as governments when it comes to the goods and services they deliver[^2].
 
+Any government contract that can only be successfully managed by another nation state or a private firm with majority state ownership should be retained, developed, administered and controlled by the UK government, until such a time as a viable private alternative becomes available.
+
+Any government contract that has been successfully tendered can be put on hiatus if it is found that the successful firm has become majority or wholly state owned. This will run until the either: 1) the firm returns to majority private ownership, 2) the tenure of the contract elapses. Should the contract elapse the government has the right to re-tender the asset.
+
+At completion of a government contract, should the asset be in proper working order to be sold to the private sector, the UK government must never knowingly undersell the asset. If an asset is sold undervalued the government minister responsible should be held accountable on condition of treason for knowingly betraying one's country. We believe that selling our nations assets at below market value is to steal from all of us and requires a severe punishment.
+
 ## Public Sector
 
 In order to stimulate adoption of the living wage, and act as a good example, all public sector organisations and [National Infrastructure Organisations](infrastructure.html) will be required to pay employees at least a living wage.
--- a/economy.md
+++ b/economy.md
@@ -71,7 +71,7 @@ Ensure worker-owned cooperatives and mutuals are given priority with all governm
 
 All private companies who receive public money must be subject to the same transparency requirements as governments when it comes to the goods and services they deliver[^2].
 
-Any government contract that can only be successfully managed by another nation state or a private firm with majority state ownership should be retained, developed, administered and controlled by the UK government, until such a time as a viable private alternative becomes available.
+Any government contract that can only be successfully managed by another nation state or a private firm with majority state ownership could be retained, developed, administered and controlled by the UK government, until such a time as a viable private alternative becomes available.
 
 Any government contract that has been successfully tendered can be put on hiatus if it is found that the successful firm has become majority or wholly state owned. This will run until the either: 1) the firm returns to majority private ownership, 2) the tenure of the contract elapses. Should the contract elapse the government has the right to re-tender the asset.
 
--- a/economy.md
+++ b/economy.md
@@ -75,7 +75,7 @@ Any government contract that can only be successfully managed by another nation 
 
 Any government contract that has been successfully tendered can be put on hiatus if it is found that the successful firm has become majority or wholly state owned. This will run until the either: 1) the firm returns to majority private ownership, 2) the tenure of the contract elapses. Should the contract elapse the government has the right to re-tender the asset.
 
-At completion of a government contract, should the asset be in proper working order to be sold to the private sector, the UK government must never knowingly undersell the asset. If an asset is sold undervalued the government minister responsible should be held accountable on condition of treason for knowingly betraying one's country. We believe that selling our nations assets at below market value is to steal from all of us and requires a severe punishment.
+At completion of a government contract, should the asset be in proper working order to be sold to the private sector, the UK government must never knowingly undersell the asset.
 
 ## Public Sector
 


OliverJackson1:
Removed the statement regarding treason, on reflection, too strong for a manifesto.

andrewdwilliams:
I think this looks okay. I agree with the removal of the treason part. I will give this a :+1: but will be interested to see what others think.

tmtmtmtm:
I'm afraid I'm finding the text here fairly incomprehensible, to the extent I'm not entirely confident I can either discern what problem it's trying to address, or what the proposed solution is. But on a face-value reading, this seems like one of the most ridiculously extreme forms of isolationist and protectionist policy imaginable, whereby (for example) if life-saving medical services had been developed by a foreign country, the NHS couldn't use them until it managed to create its own equivalent (even if that took 50 years to happen). If that's _not_ what it's saying (which I would assume that it's not, as that would be absurd), then it needs to be expressed in a much better way. Until then, 👎  from me.

OliverJackson1:
@tmtmtmtm thank you for the feedback. I consider the issue to be that occasionally for government contracts the successful tender comes from another government or private firm with majority state control. I believe that we should be pursuing an economic strategy that either 1) empowers the private sector (not one proped up by state funding) or 2) establishes the state as the majority provider. To empower a seperate state at the expense of our own citizenry is absurd. 

Would I be right in assuming point 2 and 3 are acceptable? 3. Never under sold and 2. "Can" be put on haitus depending on the situation. 

If I amend the first point to "could be considered for" development etc intra-state would that be more palatable?

Have I clarified the point sufficiently? Let me know if not and I'll give it another go. I think this is a very self-defeating hypocritical position from our government to sell our assets to other governments. For other governments to make profit. In terms of wording, could you provide more feedback on the aspects of the wording? Too many, not enough, too elaborate, etc.

tmtmtmtm:
I'm afraid I fundamentally disagree with the concept. If another state is the best provider of a service, then we should wholeheartedly accept that, just as we would presumably want them to do likewise in cases where the UK would be the best provider.

OliverJackson1:
Do you think that we should attempt to be able to provide the asset of the contract ourselves?  At any point? If it were to be amended from a retraction of service to a determination to provide, would that be more acceptable. 

As a side did you have more detail on the wording. Obviously I'm not aiming for incomprehensible. :)

OliverJackson1:
@tmtmtmtm the text in the first point is now 'could be'. 

As this opens consultation only and is non-committal can I assume this meets your approval?

In terms of fundamental disagreement, I'm not sure what you fundamentally disagree with, please elaborate. 

Also please can you elaborate on it being incomprehensible, that's something id like to avoid.

Finally, please can you respond to points 2 and 3. Are you comfortable with these?

tmtmtmtm:
I can't comment on how to improve the wording, as I'm still far from sure that I even understand what you're trying to say. What you _seem_ to be saying is so insane to me that I can't quite bring myself to believe that you're really suggesting it. Point two seems equally confused, and bizarre. I'm afraid I don't have the time at the minute to engage deeply enough on this — perhaps if some other people joined in the discussion I'd discover that I'm simply misreading these.

Point three seems to carry some odd presumptions (e.g. that every contract leads to an asset that can be sold), and is again quite oddly worded, but also feels like it's addressing some underlying issue in a little too oblique a manner. The NAO report on the sale of Eurostar — https://www.nao.org.uk/report/the-sale-of-eurostar/ — is probably a good starting point for policy adjustments in this area

OliverJackson1:
@tmtmtmtm that is very non-constructive response verging on rude. To refer to the motion as "insane" but then displaying no reasoning and rational is very bizarre to me.

What is insane about ensuring that contracts given by the UK government do not transfer power over assests or resources to other national governments? 

In terms of your original protectionist point, the other governments are acting in a protectionist way by not allowing the market to act freely on the businesses they own or control. By NOT tackling that we are green lighting protectionism, not the other way round as you assert. In terms of isolationism, there is no issue with Private MNC's, such as EON but companies like EDF I oppose. One is solely market owned the other is majority state owned transferring control of 20% of the energy generation assets of the UK to a foreign government. 

Point two seeks to protect the UK government should a private firm move into the hands of its respective government. Again ensuring the security of the contract and its contents.

Point three seeks to ensure that a fair market valuation is given for any asset or resource contained in a contract that subsequently is sold to the private market. 

I find your block here ill consider and poorly thought out with very little reasoning or consideration, to the point where you openly suggest that you don't understand it, why would you block something that you don't understand? I too welcome the opinions of others on this.

Floppy:
I've not been keeping up with this one yet, but I will take a look as soon as I can and read through the various discussions properly before wading in. Thanks everyone for the contributions so far, let's keep it civil and keep iterating towards something that works.

tmtmtmtm:
@OliverJackson1 you seem to have misread what I wrote. My block is _because_ I'm unable to understand the proposal. If you can rewrite it to be more understandable, then perhaps we can discuss it in a more considered way. (As I suggested earlier, making it more explicit what the problem is you're trying to address might help.)

philipjohn:
Please correct me if I'm wrong but I'm reading this proposal as based on an underlying assumption that government contracts operated by entities owned by foreign governments are bad.

That seems to be the bit that @tmtmtmtm is having an issue with, from what I can tell.

If that underlying assumption is where this has come from then I also have to disagree with the premise. Who operates a government contract should be of little consequence - what matters is that the contract is fair, and represents a good deal for the taxpayer.

If a French state-owned company represents the best deal for the taxpayer, then that's the best thing for the country. We should absolutely make sure that government contract rules are good enough to make sure that contracts are a good deal. We should also make sure that if government fails to get a good deal, that they are adequately held to account and that contracts can be annulled or revised.

OliverJackson1:
@philipjohn I disagree with the "Who operates a government contract should be of little consequence - what matters is that the contract is fair, and represents a good deal for the taxpayer." I personally feel that its inconsistent with both theoretical economics and national security. National assets or contract contents can be manipulated and this posses a risk. Its a self defeating economic position to disseminate limited/reduced state ownership and then push billions towards other governments.

An example, the recent 15b contract to supply Genesis Housing Association with electricity was won by four providers, of them there was Engie and Gazprom. I oppose Gazprom in the contract as there is majority control with the Russian Governement (via 3 seperate entities). There is evidence of Gazprom manipulating other nations assets to empower political negotiations. Engie on the other hand I endorse as they are internationally owned with only 30%(ish) control by the French government and 58% control via private persons.

I could go for a rewrite but I think the issue is more conceptual. @Floppy how do you retract a file change?

philipjohn:
I can certainly understand the national security aspect. That said, the world where we're scared of other governments isn't one that I want to be in, nor perpetuate, and while I'm likely to be derided as a clueless idealist, I'd rather the Manifesto were much more internationalist rather than upholding protectionism.

OliverJackson1:
@philipjohn I appreciate your note hear but I feel that I must point out one crucial point here. The countries that are controlling their private industries are being protectionist. I understand that you don't want to be protectionist but that's exactly what you are doing by trading with them. This is vital to a consistent economic approach. We must establish a clear understanding of key terms (not that you don't know them but it can often be confusing to try and align concepts of free enterprise or trade with individual companies and their interactions with states) if they are going to be used. 

Protectionism is a government action or policy that restricts or restrains international trade or investment done with the intent of protecting local business from foreign competition. If you consider that this isn't the consideration or action of companies like EDF or Gazprom (even RBS, Lloyds etc) than that's where we truly disagree. 

I do not want to support protectionist policies.

I must also say, I don't have a problem with public ownership, if its needed (profit vs infrastructure investment, asset creation, etc) but there is no logical reason to allow another nation to undertake a contractual agreement with the UK government.

I'd like an actual example from any contributor if it can be provided of a wholly owned government provider to a UK government contract.

As it is, the only contract I can consider that could be wholly considered with no private investment is military, example Trident or other military provisions.

Therefore, the iteration to the manifesto seeks to establish that our government will not support protectionism or undermine national security. Equally it will ensure that asset creation is not undermined by undersell techniques often used by the UK Gov't.

Is there anything here that you may think may call for a rewrite @philipjohn, or are we of fully differing opinions. No problems either way :).

philipjohn:
> there is no logical reason to allow another nation to undertake a contractual agreement with the UK government

I think there are probably plenty of logical reasons for the government to sign contracts with other nation states.

My objection here is that the underlying assumption in this proposal that foreign-owned entities are bad for our national security and industry. That's not the case, and each instance should be judged on it's merits rather than pre-judged because of their ownership.

OliverJackson1:
@philipjohn this does feel like an impasse. 

I would just state that 'foreign owned' is not in question, 'foreign nationally owned' is in question.

I want to see a freely trading market with no state intervention where it isn't needed. If it is needed then it isn't a free or fair market. Therefore, I don't support it and feel that our government should provide it, if it has to. 

Free markets with a proportionality of equality are wonderful subversions of power and generate societal growth.

philipjohn:

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/390), and results are set in the merge status checks below.

## Changes

PR #396, opened 2016-07-04
Amendment to Drug Wording

OliverJackson1:
Amendment from Legalisation to Decriminalisation. Then a few more words on what we'd do and the desired outcome.

### DIFF
--- a/crime.md
+++ b/crime.md
@@ -19,7 +19,7 @@ Ensure euthanasia is permissible and available for those with conditions causing
 
 ## War on Drugs
 
-We will end the unwinnable war on drugs. Recreational drug use will be legalised under a program of regulation.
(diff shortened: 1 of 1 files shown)


philipjohn:
I like this. One issue: it's not clear what "fines for improper use" means. Do you envisage there being roll is about where drugs can be used, and if so, what do those look like?

OliverJackson1:
@philipjohn, yeah, fines would be given for a number of usage breeches. This is the method being used in Portugal and is proving very effective. Essentially, you develop clear user spots (shops, clinics etc) and then use pejorative measures to limit usage in public places such as parks. Its that subtle difference between legalisation and managed decriminalisation. 

Obviously, the level of managament is different depending on the drug in question. Does that clarify?

philipjohn:
It does, thanks! I'm wondering how we can get that succinctly into the Manifesto so that others will get the idea first time. Not sure... need some thought.

Floppy:
👍 from me and can edit later for clarity.

Incidentally, that sort of method is presumably the same for alcohol, tobacco, etc now? We have fines for "improper use" in public places, right?

philipjohn:
:+1: to reflect my earlier support

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/396), and results are set in the merge status checks below.

## Changes

PR #399, opened 2016-07-04
Update and Amendment to Society

OliverJackson1:
Amended some of the spelling in the 'Never-ending' section, as well as adding a concept piece on the offering of an hours reduction vs a pay increase.

Also, amended the Egalitarianism section, to have an Egalitarianism section that doesn't ensure all humans have equal social standing is a bit off.

### DIFF
--- a/society.md
+++ b/society.md
@@ -56,12 +56,15 @@ Under all devolved jurisdications each party to the civil partnership must be of
 
 ## Never-ending Consumption and Consumerism
 
-Our entire society and economy is centred around the principle of continued economic growth driven by ever increasing consumption. Everyone in society is put under pressure to put in more hours at work, to aim for the next promotion, to get a better salary, to get a bigger house and fill it with more things - despite the fact this rarely leads to increased happiness or contentment. We feel pressure to do so because it's expected of us, because of peer and family pressure, and because of the constant stream of advertising on the web, TV, radio and billboards. The Government encourage this in order to grow the economy forever more, whilst ignoring the fact it creates feelings of peer inadequacy, is causing chronic physical and mental exhaustion in our workforce, is fuelling an obesity crisis, and is hugely damaging to our natural environment.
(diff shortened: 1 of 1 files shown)


andrewdwilliams:
:+1:

philipjohn:
:hand: I like the egalitarianism bit but I'm not convinced government should be micromanaging something like pay reviews.

@OliverJackson1 Also bit of feedback: try and limit changes to just one thing. Trying debate two changes at a time is difficult :) so we risk having both fail when one might really pass.

OliverJackson1:
@philipjohn good point, had a couple of things to contribute and didn't want to lose them. 

Not sure what you mean on the pay review micro-management, could you clarify? I'm offering an alternative to increasing pay, stating that businesses through legislation should offer either reduced hours or increased pay. The employee can then decide what to do. Outside of the Bill, there's no interference from state,

philipjohn:
We perhaps need to clarify what is meant by "pay cycle" here, and what actually happens at the end of that? My assumption was that this meant the common practice of doing annual pay reviews.

My reluctance is based on pay reviews being something that not all employers do, and each employer does them differently. In order to introduce this we'd have to effectively legislate that businesses will have to do some sort of review in order to make the offer of increased pay/reduced hours.

Perhaps you could explain how you envisage this working in practice?

Floppy:
👍 but I think separating the working conditions bit into a separate PR might be the best idea here. I have proposals on changes to pay, conditions, working week etc, coming soon as well, so I'll adjust this to suit when those come in if it has passed.

philipjohn:

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/399), and results are set in the merge status checks below.

## Changes

PR #400, opened 2016-07-04
Amendment to Economy

OliverJackson1:
Addition of Employee Equality

A few notes on actions to increase equality in the workplace.

### DIFF
--- a/economy.md
+++ b/economy.md
@@ -95,6 +95,14 @@ Follow recommendations regarding transparency in the [Lyons Housing Review](http
 information to the public in a similar manner as the property price paid data set and make it a legal
 requirement to register land option agreements, transactions and prices."
 
+## Employee Equality
+
+We wish to ensure that all employees are treated fairly and equally. 
+
+We believe that equitable treatment for Maternity and Paternity leave would help. We seek to establish paternity leave that matches the 52 weeks leave provided for Maternity cover. Equally we will pursue that organisations that offer their own Maternity leave must off the same agreement for Paternity.
+
+We believe that alongside Gender Pay Reporting, organisations should undertake Race Pay Reporting, Age Pay Reporting, LGBT Pay Reporting and Faith Pay Reporting. In line with Gender Pay Reporting organisations must publish their internal findings to the public on a forum of their choice. Findings will also be considered by Government organisations if potential discrimination is found.
+
 [^1]: [EU Financial Transaction Tax - Wikipedia](https://en.wikipedia.org/wiki/European_Union_financial_transaction_tax)
 [^2]: [Stop Secret Contracts](http://stopsecretcontracts.org/)
 [^3]: [Land Value Taxation Campaign - What is LVT?](http://www.landvaluetax.org/what-is-lvt/)


andrewdwilliams:
I'm not sure about the whole paternity/maternity leave thing. Call me naive, but I can see why mothers would want more time off - for a start, they have the physical burden of carrying the child and then birthing it, as well as feeding it. The fathers do not have that burden. Paternity leave perhaps is a good idea, but giving them a full 52 weeks may be excessive.

Also, I've never heard about pay discrimination on sexuality or religious lines before. Enforcing unnecessary pay reporting is a burden on businesses, especially in terms of cost, but also in terms of time taken up by collecting this data. If there is no actual purpose to it, it seems like quite a wasteful proposal.

philipjohn:
I'm not sure about the reporting bit, but I'm well behind the equal parental leave. :Hand:

OliverJackson1:
@andrewdwilliams the fact that mothers are in a position of need might suggest that partners should be there to support. Equally, the period after birth can be very isolated and lonely, leading to many psychological and potentially damaging effects. So I take you point but actually feel it makes the case for the equality of leave.

In terms of the length of time, equality is equality. If it isn't it isn't. I don't mean to be flippant with that, I just mean, if its not equal, its not equal.

@philipjohn are you looking away from gender reporting etc? Does your issue concern positive discrimination? Or is it the extension of reporting that is the issue? Just looking to clarify before I retort.

Following this amendment I was hoping to add another in for the rewarding of businesses that value roles instead of people. Meaning those hired can't negotiate (a practice in place with the likes of google) that tackles pay differences without positive discrimination. It toes the thin line between stopping sub-conscious bias and enforcing positive bias.

There may be an elephant in the text here, potentially, we might want to re-legislate that it is Partner Leave as opposed to Paternity Leave. This is the 21st century and many families aren't built on such binary concepts of family. 

Let me know your thoughts, happy to debate :).

philipjohn:
It's the extension of reporting bit that I'm concerned with. I'm a big fan of the state leaving people alone to get on with things unless absolutely necessary :) That's not to stay I'm a subscriber to the small state philosophy - I believe the state should be bold and intervene where necessary but be very, very particular about where it places burdens upon others. I'm not sensing enough of a need here to justify having businesses go through the additional administrative overhead.

Ha, I nearly said it in my first comment actually but yes - I definitely favour "parental leave" over "maternity/paternity leave" :)

Floppy:
👍 from me on this; the pay reporting text could perhaps be simplified to make it feel less burdensome though. Maybe talk about a pay report including breakdowns by gender, ethnicity, disability, etc, rather than implying separate returns with Capital Letters On Them. :)

Bigger picture sidenote: I reckon we should go radical though and do the [swedish thing](http://usatoday30.usatoday.com/news/world/2008-06-18-salaries_N.htm) of publishing all pay. I'll do that as a separate PR though when I get to it...

philipjohn:

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/400), and results are set in the merge status checks below.

## Changes

PR #417, opened 2016-10-26
Expand UK airports and make them carbon neutral. 

blewis99:


Floppy:
The intention of the original wording here was to attempt to force a move to sustainable transport. As far as I know, there is no realistic option for carbon-neutral air travel on the table - unless that's changed?

blewis99:
In the original wording airport expansion would be blocked until the aviation industry is carbon neutral. This would prevent jobs from being created, cause air fares to rise (making air travel only possible for the privileged few) and could endanger the UK from being cut off from important parts of the world. It would take time for a carbon neutral fuel (probably a carbon neutral biofuel) to replace current fuel on all flights taking off or landing in the UK, I believe 2029 is a reasonable target. The changes made will avoid the consequences listed above whilst still forcing the aviation industry to move to a carbon neutral fuel.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/417), and results are set in the merge status checks below.

## Changes

PR #422, opened 2016-12-08
Replace Vehicle Excise Duty and Fuel Duty

Xyleneb:
How is the wording? (is it clear and succinct?)

Also, is the rationale agreeable?

### DIFF
--- a/transport.md
+++ b/transport.md
@@ -9,13 +9,21 @@ title: Transport
 
 No new fossil-fuel powered vehicles may be sold in the UK after 2029.
 
-## Scrap Vehicle Excise Duty
+## Replace Vehicle Excise Duty and Fuel Duty
 
-As Vehicle Excise Duty is associated with vehicle emissions, it seems fairer to tax
-the fuel directly. VED should be scrapped, and fuel duty increased to cover the reduction
-in income. Overall, the tax levied should be lower, due to the removal of the need to administer
-VED, though the tax will fall more heavily on those who use more fuel, and thus create more
-carbon emissions.
+The way our current tax system works is that it assumes that fuel bought at the pumps is going to be used for public road travel. This isn't always true. ‘Red fuel’ or ‘untaxed fuel’ provide some exceptions, but generally speaking if you want to buy petrol then you’re forced to pay a tax to maintain roads, which you might not be wearing down. Meanwhile the general public are taxed variably on emissions - even though emissions are assured from the use of petroleum.
+
+'Vehicle Excise' Duty and 'Fuel Duty' will be replaced with an 'Emissions Tax' and 'Road Strengthening Tax' respectively.
+
+These changes to the titles in the tax system are meant to reflect the given purpose of the tax, making them more understandable to the public; providing greater transparency to them about what they are being taxed for, and some degree of protection against double-taxation also.
+
+The 'Emissions Tax' will be applied at a flat rate at the pumps, and will be based upon typical combustion values in the UK and upon the estimated environmental impact that these gases create.
+
+The 'Road Strengthening Tax' will be applied at a variable rate to vehicles licensed to drive on public roads, and will be based upon vehicle type and weight. Vehicles which are heavy or typically use more of the roads (such as HGVs) will pay a lot of tax for road maintenance. Vehicles which are light or which use less of the roads (such as small family cars, or agricultural vehicles) will pay less tax.
+
+What this should do is bring the cost of fuel down, while however unfortunately bringing the cost of "road tax" up. This could lead to greater evasion and therefore decreased government transport revenue.
+
+The advantage though is that families who could not afford to buy fuel for heating, cooking or travel before would see the cost of fuel come down; as Fuel Duty, a regressive form of taxation (renamed to Emissions Tax), would be reduced.
 
 ## National Infrastructure Organisations
 
@@ -27,7 +35,7 @@ Until the aviation industry is carbon neutral, there will be no further expansio
 
 ## Cycling
 
-Commit to investing in cycling infrastructure: dedicated and separate lanes; secure parking; making traffic junctions 'cycle-safe'; encouraging shared use of pavements and paths where dedicated lanes are not feasible.
+Commit to investing in cycling infrastructure: dedicated and separate lanes; secure parking; making traffic junctions "cycle-safe"; encouraging shared use of pavements and paths where dedicated lanes are not feasible.
 
 Put in place clear and simple standards to cover cycle lane design, signage, continuity and traffic signals. Extend existing transport planning requirements to always include these new cycling standards.
 
@@ -43,7 +51,7 @@ Encourage automobile manufacturers to agree upon an open and non-proprietary sta
 
 ## Electric and Alternative Fuel Vehicles
 
-Incentivise existing fuelling stations to install charging units for electric vehicles to help create a nationwide network and overcome the perception of 'range anxiety' amongst consumers.
+Incentivise existing fuelling stations to install charging units for electric vehicles to help create a nationwide network and overcome the perception of "range anxiety" amongst consumers.
 
 Incentivise existing fuelling stations to install hydrogen fuel pumps.
 


Xyleneb:
I've proof read it and I'm not happy with it. Can someone help to fix it?

I'm sure the fundamentals are good, but I'm struggling to distill it into laymans' terms...

Floppy:
The move to an explicit emissions-based tax looks good. I'll have a read through properly and suggest wording improvements later on this evening.

philipjohn:
> generally speaking if you want to buy petrol then you’re forced to pay a tax to maintain roads

This is a common misconception. Roads are funded from general taxation, not any ring-fencing of specific tax revenue. VED and Fuel duty go into general taxation. County councils set budgets for road maintenance out of their own budgets (only part of which comes from the treasury).

So while I appreciate the thought behind this proposal, I disagree with it because it would link a tax directly to an output, and unnecessarily bind revenue to specific spending :hand: 

I think what we have at the moment - scrapping VED and increasing Fuel Duty - is great because it directly taxes emissions, which is what we want to reduce.

Xyleneb:
>This is a common misconception. Roads are funded from general taxation, not any ring-fencing of specific tax revenue. VED and Fuel duty go into general taxation. County councils set budgets for road maintenance out of their own budgets (only part of which comes from the treasury).

VED is due to be largely flat-rated and ring-fenced: http://highwaysmagazine.co.uk/chancellor-announces-new-roads-fund/

The *reason* you are taxed for VED is emissions. The *purpose* for the tax is road maintenance. These are not consistent.

>while I appreciate the thought behind this proposal, I disagree with it because it would link a tax directly to an output, and unnecessarily bind revenue to specific spending ✋

If the tax man is unbound he can rob you, double-tax you, and so forth. Take his umbrella powers away and he'll be forced merely to perform his job. It's incredibly important to know why you're being taxed. I don't want it to be hidden away in a spreadsheet somewhere.

>I think what we have at the moment - scrapping VED and increasing Fuel Duty - is great because it directly taxes emissions, which is what we want to reduce.

Fuel Duty is an important metric for the general public. Under your government, you propose to hike it up. Secondly you have no plans on how to fund road maintenance once you've abolished all fossils. I like the reduction in administration and enforcement, but I can't fathom the rest of it.

philipjohn:
> It's incredibly important to know why you're being taxed.

I'm unclear on how that is not transparent at the moment. VED is a tax on emissions - it's a tax on pollution - and one that will hopefully, one-day, bring in absolutely no revenue and can therefore be scrapped.

> Secondly you have no plans on how to fund road maintenance once you've abolished all fossils.

To do so requires having a fully-costed manifesto, which is unrealistic. Not even main opposition parties are able to get all the financial information they need out of government in order to fully cost manifesto pledges so a few chancers on the internet don't have a cats chance in hell ;)

This is partly why linking taxes to specific spending pots doesn't work. The income and expenditure can be affected by different external factors, affecting the balance and creating surpluses and deficits. If they are directly linked its harder to counteract any surplus/deficit. Instead, by putting all income into one central pot and then apportioning that money via a properly thought out budget, the right amount of money gets to where it's needed *regardless of the original source*.

Xyleneb:
>I'm unclear on how that is not transparent at the moment. VED is a tax on emissions - it's a tax on pollution - and one that will hopefully, one-day, bring in absolutely no revenue and can therefore be scrapped.

VED charges you for public road use, because of emissions, to pay for road maintenance. Fuel duty charges you for fuel purchases, because of assumed road use (though as you say it's collected as general revenue), to pay for carbon off-setting (and schools and hospitals, apparently).

The reason you're being taxed should be obvious. What it's supposed to pay for should be obvious, even if it isn't ring-fenced. The consequence of it being unclear is that the tax man takes your money and you don't understand why.

>To do so requires having a fully-costed manifesto, which is unrealistic. Not even main opposition parties are able to get all the financial information they need out of government in order to fully cost manifesto pledges so a few chancers on the internet don't have a cats chance in hell ;)

You should still try. An educated guess at costs is better than nothing. Credibility is lost on those groups who fail to provide even rudimentary budgeting.

>This is partly why linking taxes to specific spending pots doesn't work. The income and expenditure can be affected by different external factors, affecting the balance and creating surpluses and deficits. If they are directly linked its harder to counteract any surplus/deficit. Instead, by putting all income into one central pot and then apportioning that money via a properly thought out budget, the right amount of money gets to where it's needed regardless of the original source.

We're unlikely to find agreement on that. Find a relevant tax to apply to cover the deficit, instead of depending on other revenue sources. That way buying wagon-wheels won't fund wars in the middle-east, and schools won't depend on funds from tobacco revenue. The inherent unfairness of the tax system stems in large part from it.

At any rate; if we discount ring-fencing from the policy submission, is it viable and worth backing in your view (as compromise may be?)

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/422), and results are set in the merge status checks below.

## Changes

Floppy:
The ring-fencing question is mainly just coming up in the comments here, it's not really reflected in the proposal itself. I'm all for making taxes clearer on why they are being levied. The last section about emissions tax being less regressive than fuel duty, I'm not sure, because it would still fall more heavily on the poor, but I'm happy we can fix that in a followup. I'm OK with this 👍

Floppy:
Sorry for the delay @Xyleneb I meant to come back and read this through much sooner.

Floppy:
I have opinions on the ring-fencing of taxes, but I won't raise them here. It's something for another proposal I think, to expand upon the principles of taxation.

Xyleneb:
> "The ring-fencing question is mainly just coming up in the comments here, it's not really reflected in the proposal itself. I'm all for making taxes clearer on why they are being levied. The last section about emissions tax being less regressive than fuel duty, I'm not sure, because it would still fall more heavily on the poor, but I'm happy we can fix that in a followup. I'm OK with this :+1:"

Ahh right, I've just seen this. Well, I don't want to imply or mislead the public into believing that we'll ring-fence these taxes if we aren't going to. That's why I intended to re-read it and possibly re-edit it.

I've seen the new vote system by the way. Who/how many members do I need to convince on my policies? Is there a time limit for sorting these things? What are the rules?

Floppy:
@Xyleneb yes, while the voting system hasn't changed, the instructions added in each proposal are new. Any suggestions on ways they can be improved are greatly appreciated (pop something in https://github.com/openpolitics/votebot/issues/new if you have ideas).

Floppy:
Just to note, there are only about 18 days left on this before timeout. @philipjohn, what would help turn your vote into a :+1:?

philipjohn:
I'm struggling with this... The proposal includes some un-evidenced assumptions that I think it would be a mistake to base policy on.

> generally speaking if you want to buy petrol then you’re forced to pay a tax to maintain roads

I mentioned this earlier, but it isn't true. If the public perched this to be the case, that's unfortunate but policy should be based on more that false perception.

> The 'Road Strengthening Tax' will be applied at a variable rate to vehicles licensed to drive on public roads, and will be based upon vehicle type and weight.

I have a pretty great car, but I do about 8k miles per year. This proposal would have me paying more tax than my sister who has a small car but uses it everyday. That's an anecdotal example, but demonstrates that the assumption that heavier vehicles generate more wear and tear is not necessarily true. Mileage is a better indicator of road use, and so fuel consumption at the pump is more likely to tax road use fairly.

I appreciate the aim here - to make the tax system fairer - but for those reasons above I don't think this proposal achieves that aim, not gets us significantly closer.

There's also two more things I'm concerned about;

* It will likely increase costs on the haulage industry pretty dramatically
* The whole country, and it's economy, benefits from good road infrastructure so, IMO, we should all pay for it through general taxation, not just direct users.

Xyleneb:
> I mentioned this earlier, but it isn't true. If the public perceive this to be the case, that's unfortunate but policy should be based on more that false perception.

Even if you say that the purpose of the tax is so we can put out fires, catch thieves, feed the homeless and maintain public parks; _purpose_ and _reason_ are not one and the same. If they _were_ the same thing (i.e "general use") then the tax man could turn up at your door demanding money from you because you "generally did things". You could be taxed for nothing. The reason you're taxed is important. It's distinct; from it's purpose, intent, use, or eventual utility.

In summary; I cannot accept that the _reason_ for the tax is "general use". It's either emissions, or it's road use - but it's not "just for general revenue". Purpose, maybe. Not reason. If it were, you'd have King John's taxmen turning up at the houses of the poor with no reason other than "because they need some more money".

> I have a pretty big car, but I do about 8k miles per year. This proposal would have me paying more tax than my sister who has a small car but uses it everyday. That's an anecdotal example, but demonstrates that the assumption that heavier vehicles generate more wear and tear is not necessarily true. Mileage is a better indicator of road use, and so fuel consumption at the pump is more likely to tax road use fairly.

This is a good point. The reason I based it on average anticipated use rather than actual use, was because I would need odometer records/black box records/"smart motorway" spy grids in order to get an accurate gauge on actual use. In terms of the government collecting information on you I thought it should be "the less, the better", so I avoided it.

Until George Osbourne's recent changes though, you already paid more VED for a large car, regardless of how much you used it. The assumption was that you would emit many grams of carbon a year, even if it stayed on your drive.

Perhaps it should be based on power instead, but power is no more of an accurate metric on road burden/use/wear than weight is. The problem is that you *need* to pay for road repairs. You need a tax that is relevant to the people who use it, who do most of the "sin" in order to create the problem, but you can't base it on actual use without creating a dystopian state. Anticipated use is the best you're going to get.

But if you scrap VED, then you'll be left with a large budgetary hole and no planned mechanism to fill it with.

>It will likely increase costs on the haulage industry pretty dramatically

I wouldn't worry too much about that - subsidies in the form of flat taxes for "fleet vehicles" are already in place for these industries. I just think that proportionally, they do not pay enough.

>The whole country, and it's economy, benefits from good road infrastructure so, IMO, we should all pay for it through general taxation, not just direct users.

Might as well share my opinion on this while I'm here. If you haven't got a penny to your name and you use your feet to walk everywhere, it seems alien to me that I would then charge you to fix the roads. It's the reason and purpose issue, again, in a way. If you're not directly committing the sin to be taxed, I don't see why I'd come to you for payment instead of the guy doing it. We all use fields indirectly, but what a farmer puts on his is not something I would expect to be held to account for.

Floppy:
This proposal has reached its maximum age, so under the rules we will close it off. However, @xyleneb, if you want to resubmit it taking into account the discussion here, please go ahead. It would be a pity for the debate here not to result in an addition :)

PR #425, opened 2016-12-16
Add equality statement to main page

Floppy:
Just realised it wasn't there!

### DIFF
--- a/index.md
+++ b/index.md
@@ -14,6 +14,7 @@ We believe that government should be:
 
 We believe that individuals should be:
 
+* Treated equally by the law and society, regardless of gender, sexuality, ethnicity, age, belief or disability.
 * Free to act in any manner that does not harm another individual, and does not infringe upon the rights of other individuals.
 * Enjoy a fundamental right to privacy from the state or their agents.
 


Floppy:
Test failures are transient network issues, no problem in this PR itself. Will fix those separately.

philipjohn:
Awesome :+1: Could we add "...or any other [thing]" replacing "thing" with something that makes sense that I can't think of right know 'cause I'm knackered :)

Floppy:
I was trying to do exactly that when I wrote it, but couldn't come up with the wording. The dictionary uses "category", does that work? I am very aware that my own status means that me choosing wording on this isn't great.

Floppy:
The wording in the [contributor covenant](http://contributor-covenant.org/version/1/4/) is probably better:

> regardless of age, body size, disability, ethnicity, gender identity and expression, level of experience, nationality, personal appearance, race, religion, or sexual identity and orientation.

But it still doesn't include a catchall.

philipjohn:
Yeah, it's difficult. It's also just occurred to me, in thinking about this again that we should consider expanding "equality" to "equality and diversity" too.

Floppy:
Well, the statement is about individual rights and the law, so equality is the relevant part here (I think?). We could have a similar statement in the previous section that a more diverse culture (with the same axes of diversity) is a stronger one. Or make sure we include it elsewhere in the manifesto.

andrewdwilliams:
:+1: Looks good to me.

Floppy:

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/425), and results are set in the merge status checks below.

## Changes

Issue #442, opened 2017-02-11
Over-arching economic/investment policy


andrewdwilliams:
So, I've been doing some looking around on an over-arching economic policy that defines the rest of our economic policies. I was wondering if anyone had any thoughts. Especially how this works with existing Infrastructure and Economic policies.

I'm yet to have a look at taxation, but I have concerns about the idea of a Financial Transaction Tax, especially with 'Brexit' looming and the (at least initial) reliance required on the financial sector.

- Re-nationalise the UK Green Investment Bank and rename and restructure it as the National Investment Bank (NIB). The NIB will also be merged with the Public Works Loan Board and the Infrastructure and Projects Authority.
- The NIB shall be capitalised with £20 billion, drawn primarily from the sale of the government’s shares of private banks following the financial crisis, the removal of [subsidies for fossil fuels](https://openpolitics.org.uk/manifesto/environment.html#subsidies), and the existing funds held by the UK Green Investment Bank.
- The aim will be to raise £500 billion in investment over the next 20 years and invest that into infrastructure and technology.
- The NIB will target particularly deprived or underfunded regions, and local authorities will also be able to make the case for investment in their region.
- Initiate a Green New Deal to develop a 21st-century infrastructure:
   - Focus on low-carbon infrastructure and technology, including public buildings, public spaces, public services and other long-term assets and facilities.
   - Launch a Public Works Agency that is funded by the government and directly employs people to work on these public works programmes.
    - The NIB shall be able to raise its own capital through borrowing and issuing bonds, and issue low-interest loans to private sector projects it deems worthy of investment. It shall, in effect, operate as a bank, rather than a fund.

An important source: https://www.publications.parliament.uk/pa/cm201011/cmselect/cmenvaud/505/505.pdf

Floppy:
This looks great, are you going to roll it into an edit on the economy page?

andrewdwilliams:
Yeah, I'll do that.

PR #443, opened 2017-02-13
National Investment Bank

Floppy:
Taking what I wrote at #442 and putting it into a proposal. It turns out this is very similar to a policy proposed by Corbyn, but that was not on purpose. I'm primarily basing the idea on the Environmental Audit Committee's 2012 report about what the Green Investment Bank should be - ie this https://www.publications.parliament.uk/pa/cm201011/cmselect/cmenvaud/505/505.pdf. If we're happy with this, we can incorporate how we would use the NIB into our already existing policies.

### DIFF
--- a/economy.md
+++ b/economy.md
@@ -4,6 +4,16 @@ title: Economy
 
 Markets are a tool for society, not as the primary reason for it. The economy should work for everyone, not just billionaires, hedge funds and investment banks. Excessive [rent-seeking](https://en.wikipedia.org/wiki/Rent-seeking) (gaining wealth from from financial speculation and intellectual property) increases inequality. In the short to medium term, we seek to rebalance the economy to reduce this by ensuring wage income is closer to rentier income.
 
+## National Investment Bank
+
+Re-nationalise the UK Green Investment Bank and rename and restructure it as the National Investment Bank (NIB). The NIB will also be merged with the Public Works Loan Board and the Infrastructure and Projects Authority.
+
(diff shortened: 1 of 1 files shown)


Floppy:
This proposal is open for discussion and voting. If you are a [contributor](https://votebot.openpolitics.org.uk/users/) to this repository (and not the proposer), you may vote on whether or not it is accepted. 

## How to vote

|--|--|--|--|
|Abstain|:hand:|`:hand:`|-1|

Votes are counted [automatically here](https://votebot.openpolitics.org.uk/proposals/443), and results are set in the merge status checks below.

## Changes

andrewdwilliams:
@Floppy For some reason this has registered as you proposing this? Not sure why.

Floppy:
oh, what the f...

right, I'll fix that. Damn. That would explain why I didn't get an email though - did you?

andrewdwilliams:
I did, yes. You can close this if you want and I'll open it again. I may have initiated this clicking through from the somethingnew.org.uk manifesto bit, so that might be something to do with it?

Floppy:
I think we should close this anyway, before anyone votes on it, as it's gone in wrong. I'll fix the code, test, and get you to resubmit. Sorry for the inconvenience!

Floppy:
@andrewdwilliams I think I've fixed the bug now, can you retry this proposal?

Floppy:
@andrewdwilliams you managed to find another bug we'd not hit before. I've fixed it now (again), so please have another go :)

This is all incredibly useful, I assure you! (sorry, though)

PR #444, opened 2017-02-15
More confident wording, and slight addition

philipjohn:
This changes the wording a bit to be more confident and also demonstrates one of the nasty things we'll get rid of.

### DIFF
--- a/democracy.md
+++ b/democracy.md
@@ -48,7 +48,7 @@ We believe in replacing the House of Lords with a more representative chamber. H
 
 ### Cease all new appointments
 
-Upon entering Government we will immediately end all new appointments to the House of Lords, swiftly ending the undemocratic selection procedure.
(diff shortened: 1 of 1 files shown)


Floppy:
:+1:

Xyleneb:
I think the constitutional reforms on the books at the moment are too deep-cutting, such that they will scare the public and reduce cross-party support. But this particular amendment isn't too bad :thumbsup:

andrewdwilliams:
:+1: also @Xyleneb if you want to open an issue on constitutional reforms that you aren't too fond on, I'm happy to get involved in that discussion.

Xyleneb:
>also @Xyleneb if you want to open an issue on constitutional reforms that you aren't too fond on, I'm happy to get involved in that discussion.

Ok I'll see what I wrote down a couple years ago and compare it to what's on the books. Constitutional reforms may have already (finally) happened in the time since though and made my scribbles a bit irrelevant.

PR #447, opened 2017-02-20
National Investment Bank

andrewdwilliams:
Taking what I wrote at #442 and putting it into a proposal. It turns out this is very similar to a policy proposed by Corbyn, but that was not on purpose. I'm primarily basing the idea on the Environmental Audit Committee's 2012 report about what the Green Investment Bank should be - ie this https://www.publications.parliament.uk/pa/cm201011/cmselect/cmenvaud/505/505.pdf. If we're happy with this, we can incorporate how we would use the NIB into our already existing policies.


Floppy:
👍

philipjohn:
Looks good :+1:

PR #448, opened 2017-02-20
Abolishment of Nude Body Scanners

Xyleneb:


philipjohn:
👍

Floppy:
👍 on principle, but we should tighten up the terminology to talk about [Millimeter wave scanners](https://en.wikipedia.org/wiki/Millimeter_wave_scanner) instead of "nude body scanners". We could say something like:

> Stop the use for security screening of Millimeter wave scanners and other devices that reveal in detail the surface of the skin under clothing"

We should also link to the privacy concerns as a footnote - there are some good links in the wikipedia page, including https://en.wikipedia.org/wiki/Millimeter_wave_scanner#cite_note-11

Floppy:
Passed, improvement can be made in a separate PR.

PR #449, opened 2017-02-21
Free Passage of Big Ben, slight reorganization

Xyleneb:
The 'Democracy' page has got some inconsistencies which will need to be addressed at some point, which is why I moved a couple sections around.

### DIFF
--- a/democracy.md
+++ b/democracy.md
@@ -78,7 +78,7 @@ In the UK, our executive branch (the government) is made up of members of the le
 
 The executive should be appointed by the legislature, on an annual basis. Ministers could then be selected from all walks of life for their expertise in subject areas, management, or other desirable skills.
 
-## Free Passage of Big Ben
+## Free Passage in Big Ben and the Houses of Parliament
 
 British subjects shall pay no fee to enter parliamentary grounds.
 
--- a/democracy.md
+++ b/democracy.md
@@ -78,6 +78,10 @@ In the UK, our executive branch (the government) is made up of members of the le
 
 The executive should be appointed by the legislature, on an annual basis. Ministers could then be selected from all walks of life for their expertise in subject areas, management, or other desirable skills.
 
+## Free Passage of Big Ben
+
+British subjects shall pay no fee to enter parliamentary grounds.
+
 ## Royal Secrecy
 
 Fully include of the monarchy within the Freedom of Information Act (FOIA), so that the royal household is legally obliged to respond to information requests[^royal-secrecy].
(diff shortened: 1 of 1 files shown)


philipjohn:
👍 but can I be pedantic and ask that "big Ben" is changed to "the Palace of Westminster" please?

Xyleneb:
I thought you'd kick up a bigger stink about being called "a subject" :)

Floppy:
👍 as well, but yeah, Palace of Westminster. Otherwise people get free access to a bell :)

Xyleneb:
>as well, but yeah, Palace of Westminster. Otherwise people get free access to a bell :)

There was a reason I put "Big Ben" and not "Palace of Westminster", just as there was a reason I put "Get rid of Nude Body Scanners" instead of "Charter for the Abolition of Millimeter Wave Scanning Arrays". I don't mind if you change the wording. But I want whatever you choose to be memorable and to resonate with the public. I would like journalists to look upon the manifesto and remember the things on it, so that they can then ask the members of the other parties: "Well, why aren't *you* doing (it)?"

It's easy to overwhelm even well-studied people with an abundance of burdensome language that's difficult to keep track of (https://en.wikipedia.org/wiki/Chewbacca_defense)

I know we've got a bunch of people who're technically minded here, but outside of here we've got a bunch of people who aren't, and in a way I don't want to exclude them for that reason alone.

I suppose what I'm saying is my view is "If simple will do, use it."

andrewdwilliams:
Arguably, you're making it more confusing as Big Ben and the Houses of Parliament are not synonymous, and your wording makes it unclear whether you're referring to just the clock tower or the whole thing. Something about the word 'of' in that context also seems ungrammatical. How about renaming the section 'Free Passage in Big Ben and the Houses of Parliament'. Because Houses of Parliament is far more commonly used to refer to the buildings than Palace of Westminster.

Floppy:
Oh, yeah, "Houses of Parliament" is better than "Palace of Westminster", and I reckon everyone knows what it means. How about that @Xyleneb?

BTW, I totally agree with your point about using simple language, and we definitely want to do that, but not so simple that it ends up saying the wrong thing :)

Xyleneb:
> How about renaming the section 'Free Passage in Big Ben and the Houses of Parliament'.

Yeah I think I'll still be able to edit it, I'll find out.

Floppy:
👍

andrewdwilliams:
:+1: also a +1 for collaboration

Floppy:
Hannibal

philipjohn:
👍 🎉

PR #450, opened 2017-02-21
No Cash for Access

Xyleneb:
I hope this will be a worthwhile policy.


philipjohn:
:+1: if it's not already?

Xyleneb:
>if it's not already?
It could prove to be fairly unenforceable or "limp-wristed", because political parties regularly take donations and in criminal proceedings the burden of proof is "beyond doubt".

Floppy:
👍 definitely worth being explicit on it :)

andrewdwilliams:
:+1: always good to be explicit

PR #453, opened 2017-02-22
Shared Parental Leave

Xyleneb:
Finland does it. Finland's a very "equal" place. Supposedly they have some of the best child support services as well. I'd say it's worth trying it out isn't it?


Floppy:
:+1:

I'm kinda surprised we didn't have something in on that already.

philipjohn:
:+1:

PR #455, opened 2017-02-24
Parliamentary Catering Budget

Xyleneb:
No 1 priority for constitutional reform - cutting the number of meat pies.

### DIFF
--- a/democracy.md
+++ b/democracy.md
@@ -70,7 +70,9 @@ All persons holding elected public office will be barred from standing as a cand
 
 Allowed expenses would be reduced to only those costs incurred as part of their job; for instance, travel and accommodation.
 
-Second home allowances will be scrapped, and MPs will instead have accommodation provided through parliament. This could be in dedicated accommodation, existing hotels, or long-term leases as appropriate. All MPs would be treated equally  under this system.
+Second home allowances will be scrapped, and MPs will instead have accommodation provided through parliament. This could be in dedicated accommodation, existing hotels, or long-term leases as appropriate. All MPs would be treated equally under this system.
+
+The overall parliamentary catering budget shall be capped at a rate of five pounds per head, from then onward following in line with inflation.
 
 ## Separation of Powers
 


Floppy:
All for having a cap, but £5 per head isn't enough detail. Is that per meal? Per year? What about the rest of the civil service? Lords? What about subsidised coffee? (Portcullis house is the cheapest place to get a coffee in London, if you can get inside).

Obviously we want to go into ridiculous detail, don't get me wrong, but there might be some more elegant way this could be put.

philipjohn:
As a big lover of pie, I'm struggling to support a proposal which would limit the amount of pie....

I'm joking of course :) Sounds good, but would be good to clarify a little what the £5 is as per @Floppy's comment.

Floppy:
Until this has a bit more detail, it's a no from me. Will happily change with a bit of clarification though.

Vote: ❎

philipjohn:
Same here, but mostly so I can try out the new Votebot voting :)

Vote: ❎

Floppy:
Oh, it's definitely a bit high, for sure. Maybe such catering subsidy should be removed completely, politicians should be well paid, and expenses brought into line with what a normal company would accept. Certainly the price of a beer in the Commons bar should be in line with the rest of the country, otherwise no wonder they are out of touch.

philipjohn:
Is it worth is trying to find out more about what the current situation is and working up from there?

Autumn-Leah:
I think if the current expenditure per head can be worked out, then we average the costs of restaurants and cafes and other food-and-drink-based establishments for a meal of burger or sandwich and a cup of tea or bottle of water within a given reasonable range of parliament, work out roughly how far most MPs have to travel to get the parliament and the costs involved, and then we can talk about what expenses should be. No point in talking about reasonable parliamentary budgeting if we don't actually have the numbers to budget with.

philipjohn:
Okay, rather than an actual figure, which is going to have to change anyway because of inflation, perhaps we can have some sort of formula?

E.g., the average price of a main meal in a central London restaurant?

Not that exactly probably, but something similar that can be reasonably easily worked out by a single person spending a few hours on it each year so we're not advocating something silly like spending millions on PWC or someone to produce a report on the cost of a meal in London.

In fact, could we piggy back on how the costs/prices are decided for school meals?

philipjohn:
@xyleneb Any more thoughts on expanding this a bit so we can get it in?

Autumn-Leah:
@philpjohn maybe it should be dropped, and then the above section changed to "Allowed expenses would be reduced to only those costs incurred as part of their job; for instance, travel, food (equivalent in price to that of a meal of an average lower class worker, any additional expense would not be covered), and accommodation." from "Allowed expenses would be reduced to only those costs incurred as part of their job; for instance, travel and accommodation."

Autumn-Leah:
I'm going to abstain for now until the issue can be resolved.

Vote: 🤐

LudovicD:
That part of London is quite expensive (due to the tastes of MPs & Lords in large part), i think letting them claim with a cap of £8/day total for food, drinks & all related would probably be fair. many/most were claiming £400/month for food anyway. & yes, to normal travel/accomodation. So i vote for such a cap. (£5/day might be acceptable also, but very likely to be too belt-tightening).

PR #458, opened 2017-03-01
Storage of all campaign materials

Floppy:
We should make sure that political parties are held accountable for their messaging and advertising. There was already something here about leaflets, but I've expanded it to cover other media, particularly online advertising and viral images. There are stories going around at the moment about very very targeted advertising being used by things like the Brexit and Trump campaigns; whether this is true or not, what parties tell people should be a matter of public record.

### DIFF
--- a/elections.md
+++ b/elections.md
@@ -38,7 +38,13 @@ Prospective parliamentary candidates must have a good connection with the consti
 
 Constituency offices of MPs, which quite rightly are publicly funded, should be solely for the use of casework officers, diary assistants etc to assist the MP in dealing with constituency issues. They should not be used for campaigning purposes or for locating agents, campaign managers, or other party officials, and thereby giving a publicly funded electoral advantage to incumbents over challengers.  
 
-To improve the ability of citizens to hold politicians to account for the promises they make, the National Archives' remit for creating an archive of Government publications will be extended to political campaign material produced by political parties. This will primarily be done by collecting copies of all material distributed by Royal Mail as part of election campaigns. Political party manifestos will also be archived by the National Archives.
+The Electoral Commission should record all campaign materials produced by political parties. As well as the material, also some detail on where the material was used, and a description of the audience. This should include:
+
+ * Party manifestos
+ * Leaflets (whether delivered by Royal Mail or by hand)
+ * Billboards
+ * Online ads (including targeted custom advertising)
+ * Social media campaigns & images
 
 ## European Elections
 
--- a/elections.md
+++ b/elections.md
@@ -38,7 +38,7 @@ Prospective parliamentary candidates must have a good connection with the consti
 
 Constituency offices of MPs, which quite rightly are publicly funded, should be solely for the use of casework officers, diary assistants etc to assist the MP in dealing with constituency issues. They should not be used for campaigning purposes or for locating agents, campaign managers, or other party officials, and thereby giving a publicly funded electoral advantage to incumbents over challengers.  
 
-The Electoral Commission should record all campaign materials produced by political parties. As well as the material, also some detail on where the material was used, and a description of the audience. This should include:
+The Electoral Commission should record all campaign materials produced by political parties and publish them openly. As well as the material, also some detail on where the material was used, and a description of the audience. The materials should be submitted and published within a short timescale during the campaign, not logged months afterwards. This should include:
 
  * Party manifestos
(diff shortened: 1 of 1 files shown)


philipjohn:
Vote: ✅

Xyleneb:
The government of the day will go around fishing up the signs and posters of it's opponents.

The government of the day will get first dibs or perhaps even exclusivity on access to this "public record" data, giving them an upper-hand in what should otherwise be an equal debate.

The government of the day could alter your rules, presenting a risk of slippery-slopery: from merely collating campaign material to mandating it's submission for vetting (and pamphlet bombs exist as a result of this kind of censorship: http://editorial.designtaxi.com/news-objects01082014/2.png)

Lastly, I think it is in some ways dishonest or unfair to hold people accountable for the things they did in 1993, particularly if they were under the age of majority at the time of saying or doing this dumb stuff.

Is there anything wrong with the old system of universities, libraries or private groups using whatever initiative they have available? Besides a few campaign materials slipping through the cracks and being forgotten?

Floppy:
@Xyleneb Ah, I see that I started the description with a sentence about holding parties accountable, sorry. This is about collecting the information - whether to hold them to account will be up to others, but this is part of a healthy political discourse. The "brexit bus" is of course a primary example.

My main concern is that with ever-increasing targeted advertising, parties could be rerunning things as bad as the 1964 slogan from the Tories in Smethwick [“If you want a n***** for a neighbour, vote Labour.”](https://www.theguardian.com/world/2014/oct/15/britains-most-racist-election-smethwick-50-years-on), and we wouldn't know because only those to whom it's acceptable would see it.

Xyleneb:
>I've made the publishing explicitly open, to avoid "first access" by the government.

Ok, but you wouldn't be surprised if come election time the release of this data to the public and other competing parties is delayed during "processing".

>My main concern is that with ever-increasing targeted advertising, parties could be rerunning things as bad as the 1964 slogan from the Tories in Smethwick “If you want a n***** for a neighbour, vote Labour.”, and we wouldn't know because only those to whom it's acceptable would see it.

My main concern is that you will use this record-keeping remit not only for exposure of those acting in an official capacity, but for drag-netting, prosecution, and censorship of data by members of the public.

Floppy:
Designing a system so that there is no technical ability to submit without it being immediately public would be simple; https://electionleaflets.org basically already does this. Then the only delay would be on parties themselves not submitting quickly enough.

How do you think it would affect members of the public? It only covers campaign materials produced by political parties (though perhaps I should explicitly extend to other registered third parties under electoral law).

Floppy:
Incidentally, I intend to start doing this for all Something New's outputs; properly publish them in a structured way that shows how it can be done. Hopefully I'll be able to go back in time as well and have the full history.

Xyleneb:
>Incidentally, I intend to start doing this for all Something New's outputs; properly publish them in a structured way that shows how it can be done. Hopefully I'll be able to go back in time as well and have the full history.

Something New has outputs?

>Designing a system so that there is no technical ability to submit without it being immediately public would be simple; https://electionleaflets.org basically already does this. Then the only delay would be on parties themselves not submitting quickly enough.

>How do you think it would affect members of the public? It only covers campaign materials produced by political parties (though perhaps I should explicitly extend to other registered third parties under electoral law).

How it affects members of the public depends on how you will enforce it. There are certain things I can do as an individual: paint signs, print papers, buy advert space online/in print, etc. So at what point does my crappy drawing of the opposition's education secretary become an official Something New document? And how do you check?

Floppy:
> Something New has outputs?

Yep :)

> How it affects members of the public depends on how you will enforce it. There are certain things I can do as an individual: paint signs, print papers, buy advert space online/in print, etc. So at what point does my crappy drawing of the opposition's education secretary become an official Something New document? And how do you check?

So, under this proposal, everything you do as an individual is fine. I don't know enough about how the big parties operate to know how they authorise official materials, but I would guess for things like use of the brand, there must be signoff somewhere. Certainly as leader of a (very small) party, I'd want us to have signoff on anything that used the official brand.

Floppy:
There's already plenty of regulation around who can spend what on behalf of whom when it comes to parties and campaigns, so I would think this would fit into the same rough structure as that.

andrewdwilliams:
Looks good to me

Vote: ✅

PR #460, opened 2017-03-02
Cash for access link added

Floppy:
Just adding a link to clarify what "cash for access" is.


philipjohn:
I feel like there should be a more neutral source of info we could link to but I can't find one.

Vote: ✅

Autumn-Leah:
Vote: ✅

Floppy:
It was the Telegraph or the Independent, so I went for the one I tend to dislike more, for balance ;)

PR #461, opened 2017-03-02
Fixed spelling mistakes and suggested changes to policy that would advance marriage equality

Autumn-Leah:


Floppy:
Oh, this is a really interesting one. It's immediately hit my cultural biases of "oh, that's a bit out there", but then having stopped and thought about it for a bit, I can't see why society wouldn't actually allow this if the participants want to. There are plenty of examples where this was or is allowed throughout different cultures.

It might affect wording elsewhere in the manifesto (including the leading paragraph above), but we can resolve that. Also note there is another proposal open on this paragraph; the content doesn't conflict though, so we can merge them safely later on.

So, after some introspection, it's a (slightly surprised) yes from me.

Vote: ✅

Floppy:
https://en.wikipedia.org/wiki/Legal_status_of_polygamy#International_law is interesting, in that polygamy violates the ICCPR, part of the international bill of human rights. However, it seems that this is because of its lack of equality in practice, being restricted  in most places to polygyny (more than one wife) and disallowing polyandry (more than one husband). Presumably making these equally legal would avoid that problem?

philipjohn:
Wow, I'd never even thought about this and went through similar feelings.

Vote: ✅

PR #462, opened 2017-03-05
Making democracy fairer with devolution and a proportional union parliament

Autumn-Leah:


Floppy:
Works for me

Vote: ✅

philipjohn:
We could make the wording a bit more concise but the principle is good!

Vote: ✅

andrewdwilliams:
Agree with @philipjohn, but that can be shored up at a later date.

Vote: ✅

PR #463, opened 2017-03-06
Fix footnotes in print versions

Floppy:
In print, URLs in footnotes don't get expanded. I've tweaked the structure of footnotes that are only links to show the URL explicitly, so that in print the reference can all be seen properly.

I've also updated some footnotes to point to more-primary sources, and included dates and sources where possible.

No policy changes.

### DIFF
--- a/communications.md
+++ b/communications.md
@@ -14,7 +14,7 @@ In a connected future, Internet access is a basic right. Moving the process of g
 
 Require all UK Internet Service Providers to provide IPv6 connections to home and business users by 2020[^ipv6].
 
-[^ipv6]: [Ofcom in denial over UK IPv6 failure](http://www.emilytaylor.eu/ofcom-in-denial-ipv6/)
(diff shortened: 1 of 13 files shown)


philipjohn:
Vote: ✅

Floppy:
SO MANY COMMITS. The current view doesn't work very well if there are lots of commits on the same branch with a lot of files. Sorry about that.

philipjohn:
Vote: ✅

philipjohn:
Thought: how is Vote of handling votes when new commits are made?

Floppy:
@philipjohn same as the old rules - upvotes must be reapplied. We keep downvotes. Basically maximum safety to avoid meaningless edits being used to wipe out blocks :)

Autumn-Leah:
Vote: ✅

andrewdwilliams:
Vote: ✅

PR #465, opened 2017-03-06
Incentivise electric cars used as grid backup storage

Floppy:
I've tried to explain the whole thing here, so it's a little longer than strictly necessary, but I think it's OK.


stringfellow:
Vote: ✅

philipjohn:
Vote: ✅

philipjohn:
"The government should" might sound better as "We will"

Floppy:
Thanks @philipjohn, I've updated the language to be more direct as you suggested. @stringfellow you'll need to recast your vote for the new version (lovely to see you, btw) :D

andrewdwilliams:
Vote: ✅

Floppy:
Votebot has marked this as passed, but it actually needs recast votes after my edit, so I'm not merging yet. I've filed a bug on votebot and will fix it.

philipjohn:
Ah, I was ignoring it in the proposals list 'cause it was green :)

Vote: ✅

PR #467, opened 2017-03-08
Government investment in modern networking

Autumn-Leah:


philipjohn:
Vote: ✅

Floppy:
Probably needs a little simplification and some links, but yes, happy with the idea. We can refine later.

Vote: ✅

Floppy:
This is similar to the purchase in the public interest thing that we have elsewhere, actually. Perhaps we can tie them up together in a future PR.

PR #473, opened 2017-03-08
Fair Use

Autumn-Leah:


Floppy:
Is there a reference we can add, or a little more explanation on what this would mean? I'm not familiar with the term "Fair Dealing".

Floppy:
(obviously I can look it up, just channeling the new readers' point of view :) )

Floppy:
Oh, interesting, I wasn't aware of the differences, and that "fair use" is just a US thing. https://en.wikipedia.org/wiki/Fair_dealing#United_Kingdom

I think this is a good thing, but it needs a bit more explanation of why it's in there.

Vote: ✅

Autumn-Leah:
I'll edit it to include links

philipjohn:
Vote: ✅

Xyleneb:
Vote: :white_check_mark:

Floppy:
Vote: ✅

PR #475, opened 2017-03-12
explicitly link drug, alcohol, tobacco and public health themes

BT-John:


philipjohn:
Vote: ✅

Xyleneb:

I want them to be on a stricter framework than alcohol and tobacco.

Floppy:
Vote: ✅

Floppy:
@xyleneb is there a middle ground that you're in favour of, between the current state and an alcohol/tobacco status, or are you against decriminalisation in general?

Autumn-Leah:
When you say the same regulatory framework as alcohol and tobacco, would that include the age limits? Would it be staggered limits? Would it be one hard line of reach age x and now you can? What's your justification for linking the three?

Autumn-Leah:
When you say the same regulatory framework as alcohol and tobacco, would that include the age limits? Would it be staggered limits? Would it be one hard line of reach age x and now you can? What's your justification for linking the three?

Autumn-Leah:
I'll be voting until this is clarified.

Vote: ❎

Xyleneb:
>is there a middle ground that you're in favour of, between the current state and an alcohol/tobacco status, or are you against decriminalisation in general?

If you say you'll make heroin as approachable as buying a beer then you will scare most of your electorate away. I don't think that's what BT-John had in mind. I just think their addition doesn't help to clarify it.

Decrminalisation is fine. But these drugs must exist on a still stricter framework than those already available.

BT-John:
It's the principal I want to establish here, not the detailed policies.  So let's say, there is some evidence that tobacco increases the risk of lung cancer, alcohol increases the risk of trips to A&E, or skunk increases the risk of psychosis.  Then, regulation would aim to reduce those risks, maybe by pricing, advertising, availability, strength, age limits etc.  As new evidence appears, then regulation changes.  Currently we have a regulatory system which seems to think that alcohol is good because it's used a lot by middle-class whites, tobacco is bad because it's used a lot by working-class whites, and skunk is very bad because it's used a lot by Afrocaribbeans.

Xyleneb:
>So let's say, there is some evidence that tobacco increases the risk of lung cancer, alcohol increases the risk of trips to A&E, or skunk increases the risk of psychosis. Then, regulation would aim to reduce those risks, maybe by pricing, advertising, availability, strength, age limits etc.

Pricing, advertising and age limits are a way for the government to arbitrarily decide what is good for the poor and the young. I'm pretty much only interested in availability and strength. "Gentle nudging" and "hard shoving" are the same thing from governments.

>Currently we have a regulatory system which seems to think that alcohol is good because it's used a lot by middle-class whites, tobacco is bad because it's used a lot by working-class whites, and skunk is very bad because it's used a lot by Afrocaribbeans.

To be fair to 'the establishment' beer's been around too long to ban it and 'skunk' has always had more immediately apparent side-effects than that of tobacco. It's largely a result of these historic circumstances that things are the way they are. The other motivation of course being who makes the money. But I'm digressing.

I don't see how comparing class A regulation with drink and tobacco regulation helps to clarify things in the manifesto or the eyes of the public. To me it is not clear what you intend to do and reminds me of the 'Monster raving loony party' of yesteryear. It could just be me. If everyone else says they get it, then fair enough and the yes votes will take it but I just thought it adds more confusion than it takes away.

Autumn-Leah:
As you've said you won't add the crucial details to this policy, and won't clarify it outside of the policy either, my vote will be remaining a no.

Floppy:
Note that the only change here is the mention of putting them under the same framework (not the same exact rules) as alcohol and tobacco, and to focus that regulation on public health. The general drugs policy is not being changed. Given that the policy is to decriminalise recreational drugs, having the same general framework as for other substances seems reasonable to me. I've already voted, but I will reiterate that vote - this seems harmless to add. To me it says that regulation will be by licensing authorities, rather than the police, which is already implied but not stated.

We don't generally require full details in policy submissions (see the contributing guide). It's better to start with a principle and expand iteratively from there - especially if just improving something that already exists :)

Vote: ✅

Floppy:
@xyleneb I would agree that the rules should be stricter, but those rules are licensing rules, not criminal rules (forgive the vague language, it's late). Hence the "same framework". I think that's what this addition means, anyway. 

Buying marijuana maybe shouldn't be as easy as buying a beer, because of whatever public health concerns are relevant (as mentioned here), but it should be *possible* under the right licensing. I can't comment on heroin, I don't know enough - hence the change of drug :)

PR #476, opened 2017-03-13
Be more assertive about IPv6

philipjohn:
The sentence here doesn't flow well from the previous paragraph. Adding "We will" improves that but also makes the manifesto more assertive.


Floppy:
Vote: ✅

philipjohn:
Haha you are no pleb to me @xyleneb :) Point taken - the Ofcom link also doesn't explain what IPv6 is!

Autumn-Leah:
Vote: ✅

PR #479, opened 2017-03-13
Reforming magistrate courts

Autumn-Leah:
My motivation for this policy is my utter disgust at the idea that you may merely bribe, coerce, or otherwise convince 3 people of the supremacy of conviction over not, or of the persons guilt through legitimate means or not, and thereby ruin somebodies life, whether that be temporarily or permanently is beside the point.

### DIFF
--- a/crime.md
+++ b/crime.md
@@ -53,6 +53,8 @@ A single, secular oath will replace the current options for witnesses in all cou
 
 Specifically outlaw religious courts, such as Sharia Law courts, by banning their operations within and outside the legal system. Support the [One Law for All](http://www.onelawforall.org.uk/) campaign.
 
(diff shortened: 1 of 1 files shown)


philipjohn:
Vote: ✅

Xyleneb:
I'm sure there is an economic argument to be had here (and it'll cost a lot of money) but I'm not especially interested in it. Trial by jury is one of the few things we have that's worth holding on to.

Vote: :white_check_mark:

Floppy:
It should certainly always be an option.

Vote: ✅

andrewdwilliams:
I've not thought about it too much, but this appears to defeat the point of magistrates courts. Either restrict further the gravity of cases that go before magistrates courts or abolish them altogether, so that cases on go in front of trained judges (and juries where appropriate).

PR #482, opened 2017-03-16
Adjustments to sex worker protections and pornography regulations

Autumn-Leah:


Xyleneb:
If you'd separate the porn laws from the rest of the sex work I'd immediately approve the former.

The problem is the latter. You're telling me essentially, that sex workers are crying out for the government to get involved and spread their gargantuan regulatory hands all over their work?

I don't think so. I can imagine them saying "just stop prosecuting us for doing our job", which is why the solicitation law amnesties came about. But to contain and then to strictly regulate it will not work in any of the areas where you want it to work.

That there is enough reason for me to oppose, but then you've also got the NIMBY's. The over 65's who vote en masse, among whom very few of them want a brothel outside of their house...

Just let the workers work without persecution and you will see improvements across the board.

Autumn-Leah:
@Xyleneb I appreciate your frustration with my ideas, but your objections don't actually make comment on the change. The brothel thing has already passed, if you want to vote it out, by all means go and do that, but I've attempted to make the legislation fairer than it otherwise was by avoiding worker exploitation, explicitly changing the wording to legalise sex work at home, banning set-ups which would create a profit motive, an objection in the previous thread, and ensuring that nobody gets STD/STIs thereby avoiding a public health crisis the moment the hordes of people try it out on opening day.

May you please evaluate these changes on their merits rather than complaining about the policy that's already been passed?

Xyleneb:
>May you please evaluate these changes on their merits rather than complaining about the policy that's already been passed?

Ahh right.

Points of contention:
State-funded brothels
Banning the solicitation of sexual services leaves mutualistic customer/service arrangements (think the craigslist nonsense) in a legal grey area.
The point about abortion doesn't need to be made there. It'd lose more votes than gain any.

Compared to the current status quo it's an improvement and my vote will be to approve either way.
Vote: :white_check_mark:

philipjohn:
> receiving donations from taxes

I'm not keen on these effectively​ being, at least partly, state funded.

> and abortions would be provided without charge, as it is under the NHS.

Is it intended that abortions could be carried out at the licensed premises? That's how this reads to me, which I'm not keen on. If not,i think we should leave this out - it doesn't seem necessary to mention this when we're not changing anything about abortion policy.

Otherwise,I'm happy with this proposal.

Vote: ✅

Autumn-Leah:
Yeah, that section on abortion *is* rather unnecessary. The state funding is to ensure that profit motive cannot come into play regarding the wellbeing and finances of the workers.

Floppy:
This has passed, so I'm merging it. @philipjohn I suggest you make a subsequent edit on the abortion aspect.

Floppy:
Belay that, I got confused over dates. Not merging for another 3 days min, so @Autumn-Leah if you want to edit the abortion aspect then go ahead.

PR #483, opened 2017-03-16
Lords Spiritual

Autumn-Leah:


Floppy:
Hell yeah. I thought this was already in, actually.

Vote: ✅

Floppy:
This might be better in the "cut the numbers" section, as we could just remove the Lords Spiritual in that process, but it can move later.

Xyleneb:
"I think that vast constitutional changes; such as implementing a house of citizens, or abolishing every last one of the bishops, will scare the old people. It'll scare me too probably."

There's no point in vast constitutional reform that you cannot hope to implement.
It's unsatisfying compromises or it's nothing. I'm sorry, fellow republicans.

philipjohn:
I think we need a bigger "YES" button...

Vote: ✅

Autumn-Leah:
> "I think that vast constitutional changes; such as implementing a house of citizens, or abolishing every last one of the bishops, will scare the old people. It'll scare me too probably."

> There's no point in vast constitutional reform that you cannot hope to implement. It's unsatisfying compromises or it's nothing. I'm sorry, fellow republicans.

The way I look at it, the church and state can be entirely detached as a realistic goal, and to do so would be a positive thing. You are free to have a dissenting opinion. Being that this isn't actually a criticism of the specific change of removing the CoE Lords but moreso the nature of the change, I can't say much more than that. Unless you consider the reduction in Lord numbers a constitutional change.

Xyleneb:
>The way I look at it, the church and state can be entirely detached as a realistic goal, and to do so would be a positive thing. You are free to have a dissenting opinion. Being that this isn't actually a criticism of the specific change of removing the CoE Lords but moreso the nature of the change, I can't say much more than that. Unless you consider the reduction in Lord numbers a constitutional change.

Just to be clear I thought that the Bishops and the 'Lords Spiritual' were the same thing. I'm still not sure if they are or not.

Autumn-Leah:
https://en.wikipedia.org/wiki/Lords_Spiritual - ya could've searched it buddy, but here ya go.

Floppy:
This entire page is about massive constitutional reform. It's kind of the whole point :)

It's written as a series of reforms though, one after another, so that it doesn't propose changing everything at once. Because yes, that would be impossible.

Vote: ✅

geeksareforlife:
I am all for separation of the church and state

andrewdwilliams:
Vote: ✅

PR #485, opened 2017-03-16
New page for gender equality and a summary of our current position

Autumn-Leah:


Autumn-Leah:
See: https://votebot.openpolitics.org.uk/proposals/484 for the other change that pairs with this one

Floppy:
This is an area that could be expanded a lot, so yes, a new page could well be justified. Should we, however, make it cover all aspects of equality, as there will be many? 

On the wording, I'm wary of voting on this now, as I'm very much part of the problem when it comes to equality and representation, so I'll do some careful research before casting my vote. For instance, I'm not sure if at the moment this precludes positive discrimination; now I'm not saying we *want* that, but I don't want to vote until I'm sure :)

Thanks for taking the initiative to expand this section though, it's a great idea. AND, it proved my "new file" button works ;)

Autumn-Leah:
> This is an area that could be expanded a lot, so yes, a new page could well be justified. Should we, however, make it cover all aspects of equality, as there will be many?

My fear is that the policy may get so dense in this area that separate pages will be needed anyway, so I figure that it will be better to make the pages beforehand, rather than having to detangle a long page into several pages after the fact. I'm starting with gender equality since we actually already have one (albeit badly written) policy on it, which I've summarised and transferred to here.

> On the wording, I'm wary of voting on this now, as I'm very much part of the problem when it comes to equality and representation, so I'll do some careful research before casting my vote. For instance, I'm not sure if at the moment this precludes positive discrimination; now I'm not saying we want that, but I don't want to vote until I'm sure :)

I could go on a long rant about how positive discrimination is a crap deal for everyone involved, but given that the policy specifies egalitarianism and the *supposed* goal of "positive discrimination" efforts like affirmative action programs and gender specific scholarships is to achieve egalitarianism. So yes, and no, depending on how you see it. What Egalitarianism means to us will be defined by us, and the moment it doesn't, we counter campaign or leave, which is how literally everything works in this party.

TL;DR: The party will decide whether it does or doesn't, as they've been doing so far. It's not an issue.

> Thanks for taking the initiative to expand this section though, it's a great idea. AND, it proved my "new file" button works ;)

No problem : D

Floppy:
I think it's probably better to start with a page covering a wider area, and then split down when it gets big enough - that way we don't end up with a bunch of nearly-empty pages until someone writes enough things. But yes, the idea is fine.

Vote: ✅

philipjohn:
Vote: ✅

PR #487, opened 2017-03-16
Adjustments to the House of Citizens policy

Autumn-Leah:


philipjohn:
Could you explain a little your reasoning for explicitly excluding the CofE, and does that just apply to high ranking CofE people?

Autumn-Leah:
Purely on the basis of separation of church and state, in line with the Something New's push for more secularisation in the legal and political systems.

Autumn-Leah:
It would only apply to the higher ranks as specified in the policy, a local vicar should not be limited from office as they have very little power in the church and its institutions.

Floppy:
C of E should probably be broadened to other religions too, though I'm not sure how you can word that without covering all faith groups and casting the net too wide. But I'm in agreement, generally. "High ranks" would need defining, but that's maybe a matter of details down the road.

Vote: ✅

philipjohn:
Thanks for the clarification! I agree with @floppy on broadening that bit out and agree with the general policy.

Vote: ✅

Floppy:
This has passed, so will be merged, but the C of E wording should probably be addressed in a followup proposal

PR #489, opened 2017-03-16
Changes to school uniform regulations

Autumn-Leah:
The only intent of most of my policies is to increase freedom both individually and collectively, I think this policy does this successfully.

### DIFF
--- a/education.md
+++ b/education.md
@@ -8,7 +8,15 @@ We believe that education is the keystone that makes all other longer-term strat
 
 Funding for Sure Start children's centres is to be restored, and any closed children's centres are to be reopened.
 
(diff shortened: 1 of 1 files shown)


Floppy:
I'm going to abstain on this, as I know my experience of what helps kids learn is absolutely zero, so I can't say whether this is a good thing or not. I'm all for freedom and individual expression of course, but I'm just not an expert here.

Autumn-Leah:
That's fair, I just hope this passes because individual expression and equality was the goal of this policy, whilst maintaining classroom order, for obvious reasons.

philipjohn:
You've managed to take something about which I get very passionate, apparently read my mind, but then actually articulate it in a way that isn't a shouty rant. Kudos :)

Vote: ✅

Floppy:
I think I can now properly abstain...

Vote: 🤐

PR #490, opened 2017-03-16
Equality in rape law

Autumn-Leah:


Xyleneb:
It would be easier if you would stop making lists of groups you approve of (somebody will inevitably be left out even if you extend "LGBT" to "LGBTAODFHPUSRG") and just state what rights your proposal grants to everybody, uniformly, without picking favourites.

I don't agree with that, or with using the 'PIV' abbreviation without explaining what it is (I know now, I think, having sat down and thought about it), or with confusing rape with the slightly lesser offence of sexual assault.

But if I'm to make concessions, and pretty big ones in the name of compromise (such as the policy to disarm our nuclear programme) then as these go they're pretty small disagreements.

This submission I think attempts to extend rights to all, rather than to limit them for some or to take them away... and that seems to be rare in this party/experiment.

On reflection I have to say go for it

Vote: :white_check_mark:

Autumn-Leah:
LGBTQIAP is Lesbian Gay Bi Trans Queer Intersex Asexual Pansexual + which covers all major groups other than non binary gender stuff, the plus is any other gender or sexuality stuff you believe in. The commonly accepted ones go in the acronym, I'm not just fucking about with the alphabet as you seem to be implying. It also doesn't "leave anyone out", that's the purpose of the plus.

PIV means penis-in-vagina, a commonly used term and acronym.

Anyways, thanks for approving! Proposals section is a little dead for the last 2 days so I'm glad to see someone consider my proposals today.

Xyleneb:
>The commonly accepted ones

Are there unaccepted ones?
Can you see the identity-politic can of worms that's easily opened up by this?

>It also doesn't "leave anyone out", that's the purpose of the plus.

You could have just wrote gay+. Or just plus. It encompasses all, I'm told. I've also been told it's not really an issue that's worth debating, but then all of something new's policies read like arbitrary lists like that. At some point you have to ask the question whether the party represents you or the 2% of the population that are gay - then you cast your vote one way or another.

You do not have to write lists or "explicitly include" me or anyone else. Build good things and they will come.

Autumn-Leah:
No, there's no unaccepted ones, that wouldn't make any sense. Groups of people have words for themselves, not words that *aren't* themselves. The only concession I can really make on this is that "commonly accepted" is vague af. Regardless, the point of that section is to make the point that it discriminates against these groups...which it does. Your objection is that I used a phrase that you didn't like, not that you didn't like the policy, I'm done addressing this objection.

>  party represents you or the 2% of the population that are gay 

The party represents whoever feels they are represented by it (in some contexts), the party members and/or contributors (in some contexts), and/or the constituents of Something New politicians (in other contexts).

philipjohn:
I agree with @xyleneb :) We really should expand the first use of PIV and have the abbreviation in brackets afterwards so as to educate the reader and so the next use of PIV makes sense.

Autumn-Leah:
That good enough or do you have other suggestions?

Floppy:
Vote: ✅

Autumn-Leah:
@philipjohn Planning on voting?

philipjohn:
Looking at the [existing legislation](https://en.m.wikipedia.org/wiki/Rape_in_English_law), I'm not sure if this is quite right.

> we will have it encompass oral sex

And the legislation says;

> penetrate the vagina, anus or mouth of another person

Which seems to cover oral sex already.

> all common forms of sex that can be performed by 2 women, all common forms of sex that can be performed by two men

This bit seems to be covered by

> with an object or penis

I guess you could argue changing that to "object or appendage" instead.

> forced envelopment (female on male rape)

This one might be onto something. The legislation seems to say that the offence is only committed by the person doing the penetration, so a woman forcing a man to penetrate here would not be committing rape by the looks of it.

If so, we should definitely broaden the offence to forced​ penetration by anyone. That would also cover female-on-female rape.

Vote: 🤐

Autumn-Leah:
> 1 Rape
(1)A person (A) commits an offence if—
(a)he intentionally penetrates the vagina, anus or mouth of another person (B) with his penis,
(b)B does not consent to the penetration, and
(c)A does not reasonably believe that B consents.
(2)Whether a belief is reasonable is to be determined having regard to all the circumstances, including any steps A has taken to ascertain whether B consents.
(3)Sections 75 and 76 apply to an offence under this section.
(4)A person guilty of an offence under this section is liable, on conviction on indictment, to imprisonment for life.

Autumn-Leah:
@phillipjohn Instead of citing wikipedia, if you care to actually look at the legislation that said you looked at, it specifies a *male perpetrator* meaning that *no other gender can be perpetrators of rape under current law*. If you'd like to challenge a proposal to amend legislation, it's a good idea to actually go read that legislation first. That section if you look in the Contents page of the Sexual Offences Act 2003 is just at the top under the category "Rape".

PR #491, opened 2017-03-16
Reducing the carbon emissions of taxis

Autumn-Leah:


andrewdwilliams:
Vote: ✅

Floppy:
Vote: ✅

Xyleneb:
Tax breaks in exchange for selling your data to the government is extortion by the backdoor.

Autumn-Leah:
I'm anti-government-surveillance too (tbh I'm also anti-government), but businesses already consent to these kind of checks by the police if they believe something suspicious to be going on, and the government obviously checks over your income, so I don't see incentivising both financial and environmental accountability through this method as a bad thing, especially as it wouldn't be mandatory.

Xyleneb:
>I'm anti-government-surveillance too (tbh I'm also anti-government), but businesses already consent to these kind of checks by the police if they believe something suspicious to be going on, and the government obviously checks over your income, so I don't see incentivising both financial and environmental accountability through this method as a bad thing, especially as it wouldn't be mandatory.

Imagine being a cab driver. Your choice is to sell your freedom for those rebates, or to not be competitive.

I would support this policy for one reason alone: tax rebates for the workers of those companies, themselves.

philipjohn:
Vote: ✅

Autumn-Leah:
@Xyleneb Whilst I would support that, the reality is that the workers aren't the ones buying the cars for the company. Unless you're coming at this from the angle of the workers applying bottom up pressure for better cars due to the tax breaks that they'd be getting, but that doesn't really incentivise the businesses itself to do so unless they start striking or something.

I'm just trying to figure out how that would fit into the policy, because bottom up pressure would also help but wouldn't have as much effect unless you're proposing that both a business and worker exemption be legislated into the tax system.

PR #492, opened 2017-03-16
Ridesharing and Carpooling

Autumn-Leah:


Xyleneb:
>Through a series of targeted marketing campaigns and PSA's we would encourage carpooling, especially for parents and those on the same street with a shared workplace, or workplaces in reasonable walking distance of each other.

I take it some 650 MPs will be carpooling to work too?

>Ridesharing would be promoted through The Job Center

"Have you considered doing some ridesharing today?"
"Yeah thanks for that, but we're all unemployed and can't afford to run a car."

Ahh yes, the top-down finger wagging that means well...

The best I could do for the policy is abstain from the vote.

Autumn-Leah:
It's not supposed to be the sole function of the job center, I just see it as an additional thing that those with cars could be encouraged into alongside traditional jobs.

philipjohn:
Could we expand PSA please? Acronyms are nice but cause confusion to those who don't already understand them.

Also not overly keen on mentioning Under directly too, given they are becoming something of a Marmite (yes, i get the irony of using another brand in that sentence ;) ).

Vote: ✅

Autumn-Leah:
Expanded the acronym for you. Uber is just an example, I can put "for example" or "ridesharing services such as Uber" if you like.

Autumn-Leah:
Adjusted it for you

philipjohn:
Thank you!

Vote: ✅

Floppy:
Vote: ✅

PR #495, opened 2017-03-22
Alternative medicine

Autumn-Leah:
Pseudoscience has no place in our health service

### DIFF
--- a/health.md
+++ b/health.md
@@ -18,7 +18,9 @@ The NHS will not deny care to anyone for any reason, regardless of legal status,
 
 ## Alternative Medicine
 
-All "alternative medicine" treatments currently paid for under the NHS will be immediately defunded, subject to proof that the treatment(s) work. As these treatments currently have no scientific evidence in favour of their efficiacy, they should be considered a waste of taxpayer money, given the very real potential that NHS funding is going towards treatments that don't help the patient receiving the treatment.
+Whilst the NHS remains under the full control of the government (subject to change through our NHS independence policy), all "alternative medicine" treatments (and treatments that have otherwise not been proven to work after being subject to rigorous scientific and medical analysis) currently paid for under the NHS will be immediately defunded, and subject to proof that the treatment(s) work. As these treatments currently have no scientific evidence in favour of their efficiacy, they should be considered a waste of taxpayer money, given the very real potential that NHS funding is going towards treatments that don't help the patient receiving the treatment.
+
+If the NHS is made independant (as per our policy or otherwise), we will then lobby for this to be the case, but will exert no governmental control over the independant organisation.
 
 ## Maternity
 
--- a/health.md
+++ b/health.md
@@ -16,6 +16,10 @@ NHS funding will be protected by legislation that binds the Government to commit
 
 The NHS will not deny care to anyone for any reason, regardless of legal status, economic status, or nationality (for further details, see ["Do I Treat This Immigrant?"](http://doitreatthisimmigrant.com/)).
 
+## Alternative Medicine
+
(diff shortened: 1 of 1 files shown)


Floppy:
Yep. 

Vote: ✅

tmtmtmtm:
Strongly against. This conflicts with the promise to make the NHS independent. "Politicians cannot be trusted with the NHS" applies to us too, and to believe in that must mean trusting the NHS to come to decisions like this themselves. The position must not be "Politicians must not meddle in health matters … unless it's for things that we support". :negative_squared_cross_mark:

Autumn-Leah:
@tmtmtmtmt Do you want me to change it so it states that this would only be the case whilst the state was still controlling the NHS? Either way, surely we'd have no power over it (and similarly neither would this legislation) anyway once it went independent, gives that this proposal deals solely with funding, and mentions that said funding is taken from taxes?

Autumn-Leah:
*reads the manifesto again*

...oh

Either way, the first question still stands, ignore the second one.

philipjohn:
I agree with @tmtmtmtm on this. There is perhaps a way to achieve your aims with a different policy though. Something like; "Until the NHS Constitution is in place, we will remove NHS funding for any unproven treatments. We will then lobby for the NHS Constitution to ban unproven treatments."

The wording isn't thought through (I don't like using "ban" for instance) so needs changing but perhaps something similar achieves the same aims while preserving the independence and asserting our vision?

Vote: ❎

Floppy:
Yeah, OK, you guys are right. I totally think the outcome is correct, but it's inconsistent with the self-determination we want. I think this conflict was raised before in https://github.com/openpolitics/manifesto/pull/363, on homeopathy on the NHS.

I would love to find a way that allows both. Perhaps something like "public funds provided to the NHS shall not be spent on treatments without scientific validity"? Does even that conflict?

I also like @philipjohn's wording; there are other places in the manifesto where we talk about short and long-term change, so this could be a short term thing taking effect immediately before the long-term self-determination can be executed.

I'll cheekily leave my vote where it is for now, as I'd really love to find a middle ground that works :)

Autumn-Leah:
That ok? @tmtmtmtm @floppy @philipjohn

philipjohn:
Vote: ✅

Floppy:
Vote: ✅

Xyleneb:
I want that skill you have to reshape this stuff in the face of complex problems and rattle it out so quickly. It took me a good 20 minutes just on how to put the compliment.

I can't oppose this. Vote: :white_check_mark:

Autumn-Leah:
@Xyleneb Thanks for both making me laugh and for the compliment :D

tmtmtmtm:
I'm afraid I still have to leave my :negative_squared_cross_mark: on this. At its heart this is still simply political interference in health matters, and its inclusion substantially weakens our claim that that is a Bad Thing. What treatments are available to patients should be decided by the medical community, not by politicians. 

According to http://www.nhs.uk/Livewell/complementary-alternative-medicine/Pages/complementary-alternative-medicines.aspx "The availability of CAMs on the NHS is limited, and in most cases the NHS will not offer such treatments", which makes me question how much of an issue this even is, but if we want to signal something along the lines of the this proposal, then I'd be more inclined to back something along the lines of: "We support the conclusions of (suitable medical body) that (whatever their conclusions about complementary and alternative medicine on the NHS actually are). Our policy is to ensure that such decision-making powers rest with the medical community, rather than politicians, but until then we would support introducing this through legislation."

Autumn-Leah:
@tmtmtmtm but that's exactly what this says in different words.

tmtmtmtm:
@Autumn-Leah it says the second sentence, but the first is notably absent. The proposal as stands basically says "We are going to impose our personal beliefs about medical treatment on everyone, whether the medical community agrees or not", even though it's our position that politicians doing exactly that is a very bad thing. I am only willing to back a proposal that starts from a statement that this is what the medical community _does_ want. If it does, then this should presumably be really easy to add, and the proposal would be much stronger for it.

Floppy:
As it stands, this has passed the vote, but of course we can continue adjusting in further proposals.

PR #496, opened 2017-03-22
Media classification

Autumn-Leah:
This is a big one. Representation in media is an important way in which peoples perceptions are shaped, thus having ratings based on these things makes sense, these would not be in any way legally binding and age restrictions would be done away with for the reasons stated in the policy. I'd explain further about the BBFC but I think I covered it rather thoroughly.


Floppy:
This is an interesting one. I'm thinking. :)

Floppy:
(thinking out loud) Is it that they *must* have such ratings, or that the opening up of the BBFC would allow/encourage such ratings to be created and adopted, a bit like a fairtrade food stamp? I'm all for equality and improving representation in media, but a *compulsory* rating system across such things seems draconian.

Floppy:
*For now* I'm a no on this, is feels rather too nanny-statey. But I like the idea of allowing it or encouraging it to happen naturally.

Vote: ❎

Autumn-Leah:
The movies don't have to be rated at all, but if they are to be rated, they must have all ratings present on the casing.

Autumn-Leah:
It doesn't regulate the customer, only the rating org, so that people know what movies they're buying. It's a consumer advice measure, not a measure to restrict the liberties of individuals.

Autumn-Leah:
They would all be separate ratings, not one joint rating.

Xyleneb:
I don't know what the law is on this. What separates commercial publishing from youtube?

Autumn-Leah:
@Xyleneb One's on YouTube and one's in IRL stores. Thankfully the policy doesn't deal with that, all it does is add mandatory ratings that each rating org must provide for it's films/customers (depending on whether it's a for profit NGO or not) and thereby encouraging consumers to be more informed, ensures that age ratings aren't binding, and allows movies to not be rated at all if they don't want to be, since the power removal from the BBFC would ensure that they have little to no power to ensure films and other media get rated in the first place (unless there's legislation in the way, in which case tell me so I can add the repeal of it).

geeksareforlife:
I really like the idea of having additional classifications, but I can't get past having minimum ages for buying films.

As a parent, I rely a fair deal on these (only U/PG at the moment, but I am already looking at the difference between 12/12A). I completely understand that individuals mature at different rates, but sometimes we need broad strokes.

Also, pornographic films - how are we to define those? Who, if not the BBFC, decides what a pornographic film is? I can see that being a rabbit hole that no-one really wants to go down!

PR #497, opened 2017-03-22
Housing Cooperatives

Autumn-Leah:


Xyleneb:
>We would encourage the creation of housing co-operatives by those who own their own

How will you do this?

philipjohn:
Vote: ✅

Floppy:
Vote: ✅

Autumn-Leah:
@Xyleneb If I'm honest, I'm not quite sure, I just figured I'd put it in for now and wait until I figure it out, since housing co-ops are definitely a benefit to the population. Since it involves no earth shattering moral, philosophical, or ethical issues in it, I figured being vague in this instance wouldn't hurt. Thing is with housing co-ops is that they have to be bottom up, not top down, meaning it can't be solved with legislation, so whilst I'm blanking atm, I wanted something in there as a temporary point of "this is a good thing".

Autumn-Leah:
Ad campaigns and PSA's maybe? Initiatives through unions (since people in unions are already predisposed to want to do this kind of thing)?

Floppy:
Yes, the manifesto has plenty of "work-it-out-later", so we generally don't mind that. But of course if there are good ideas to add, go for it.

Xyleneb:
I would clutter up the issues page with it. Without explaining how, it doesn't really do anything as a policy. It's like "we'll be nice to the old people". Not quite that vague perhaps. But what's important is whether you send them a card, or whether you replace their lungs with gills and then force them to go live in the sea. Without knowing the implementation it doesn't say much.

You could have tax rebates or tax funds/grants for things like rent or house purchases, but then the commercial property sector would be in uproar over the state "meddling in and biasing the market" in favour of co-ops. I'm not saying that's necessarily bad. They ought to have distributed more housing in the first place. You could cost it later too, as long as the rebates were small.

PR #499, opened 2017-03-22
Re-evaluation of policy on the PKK

Autumn-Leah:
Whilst the acts committed by the PKK are not things I would attempt to justify, the systematic erasure of Kurdish culture and lifestyles is abhorrent, and as such I believe that the Turkish government must do a lot more to both compensate for previous actions and to advance the rights of Kurdish people before it can expect any form of peaceful resolution.


philipjohn:
This sentence has a couple of grammar quirks, probably just cause they were part of the original;

> With a view to peace in the region, the PKK should it commit itself to disarming in Turkey and re-engaging in the peace process

Floppy:
Also I think there is a footnote missing, 'mp-arrested-speaking-kurdish'? But yes, good expansion.

Vote: ✅

Autumn-Leah:
@philipjohn Do you have any intention to vote on this or?

philipjohn:
Vote: ✅

PR #501, opened 2017-03-22
Repealing mass surveillance

Autumn-Leah:
I don't think I need to add much here.


Floppy:
Vote: ✅

philipjohn:
Vote: ✅

Xyleneb:
It's all gold except for the interception warrant bit. I don't wish to bar security from *targeted* interception, so long as they have obtained the applicable warrant from a judge. Can you explain what the interception warrant is and why is it's existence is contentious?

Autumn-Leah:
@Xyleneb You misread it.

>  and to prevent the existence of interception warrants and any data collected from being revealed in court.

means that *the warrant can't be revealed in court* not that interception warrants are banned or otherwise, it makes no statement on the legality of interception warrants. The policy proposal would allow these warrants to be submitted in court, which can't currently happen due to the Regulation of Investigatory Powers Act 2000, which this would repeal.

PR #502, opened 2017-03-23
Adding abstain voting

Floppy:
The votebot now supports abstentions, so they need adding here as well as an addition to the voting rules. I've enabled them in the bot already as this is really a non-issue, but if there is a problem with allowing this, I'll roll it back.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -44,12 +44,13 @@ People who have contributed to the manifesto are eligible to vote on proposals. 
 
 The simplest way to vote is using the [voting interface](https://votebot.openpolitics.org.uk/proposals). Click on the proposal details to see the change, comment, or cast a vote.
 
-Three vote types are available:
+Four vote types are available:
 
 |vote|symbol|score|
 |--|--|--|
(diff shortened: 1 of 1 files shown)


Xyleneb:
I usually just abstain by dodging the vote like Diane Abbott. What's the difference?

Autumn-Leah:
I'm tempted to abstain from this as a joke, but I won't :D Good work re: the abstain vote.

Vote: ✅

Floppy:
@xyleneb if someone didn't vote we don't know if they didn't get round to it, or meant not to. It's not compulsory at all, but a way of saying "I'm not voting". A bit like a "none of the above" in elections.

Autumn-Leah:
@Floppy Speaking of the abstain vote, you might want to add it to the vote count system on the proposals page since abstain votes don't have a column yet

Floppy:
I was wondering about that, whether it will add anything or just add more clutter to those lists. I'll see what it looks like.

philipjohn:
Vote: ✅

PR #504, opened 2017-03-26
Broaden the candidate connection

philipjohn:
Candidates for *any* elected position should have a demonstrable connection with the area they wish to represent. For local elections it's too common for the party to spread their members out over the area, resulting in councillors representing a ward they don't live or work in.


Autumn-Leah:
Yup, I agree, if we're going to have constituencies in the first place, have the elected officials actually *be from the constituencies*

Autumn-Leah:
Vote: ✅

Floppy:
Vote: ✅

PR #505, opened 2017-03-26
End co-option

philipjohn:
The electorate should always be given the opportunity to elect a new councillor if a vacancy arises on their local council.


Autumn-Leah:
Yeah, this is bullshit, thanks for the addition @philipjohn

Autumn-Leah:
Vote: ✅

Floppy:
Vote: ✅

PR #508, opened 2017-03-28
Suggested additional detail on nature of cheap, continual improvement courses.

SimonVery:


Autumn-Leah:
Both I (to vote on this), and this policy, need clarification.

philipjohn:
Thanks for the proposal @SimonVery! 

It looks like "design" should perhaps be "designed" - is that right? We can get that changed easily enough if so.

Could you elaborate on what "peer/community assessed collaborative project based" is in practice? Does this refer to the work that students would do as part of the course? Or does it perhaps refer to how the course itself is designed?

Thanks!

PR #510, opened 2017-04-06
Free school meals for all primary school children

philipjohn:


Autumn-Leah:
You've got a typo. A instead of All.

Floppy:
yes but with typo fix :)

Vote: ✅

philipjohn:
Well spotted - changed :)

Floppy:
Vote: ✅

andrewdwilliams:
Isn't there a risk that this is wasting money by subsiding families who can afford school meals, which is the majority of families?

philipjohn:
I don't think so - schools prepare meals anyway. We're getting rid of a bunch of admin for a start, we're ensuring all kids have at least one good meal per day which will have noticeable health benefits, reducing the burden on the NHS. We remove a bunch of sigma around free school meals, and the associated bullying that goes on in schools. Not at all a waste IMO.

VECTORBOY:
This has been introduced in my area and has led to schools having a half day Friday so that it is one less meal to provide. The impact being less teaching, and a higher cost in childcare.

I personally never had enough money to afford the equivalent meals that those with dinner tickets had. It should be a baseline tho, and steps should be taken to avoid waste.

I'm thinking water, fruit or such like. Reducing need of full kitchens and staff, but that miggt be too far.

But right now some kids get a roast dinner everyday and some eat nothing.

(kids will just still go to the nearest chip van)

Autumn-Leah:
To satisfy Vector and myself I'd say change this:

> All primary school children will receive free school meals, removing the stigma around only poorer children receiving free meals, and helping to ensure all primary age children are receiving at least one nutritional meal each day.

to this:

> All primary school children will receive free school meals for every standard school day (for most schools, Monday - Friday), removing the stigma around only poorer children receiving free meals, and helping to ensure all primary age children are receiving at least one nutritional meal each school day, regardless of days off, bank holidays that aren't in school breaks, or half days.

philipjohn:
> This has been introduced in my area and has led to schools having a half day Friday so that it is one less meal to provide.

That seems like a fairly ridiculous thing to do! Presumably it's the school, or the LEA, that has introduced the free schools meals, and as they largely have control over where they spend the money, they choose whether to spend it on meals or teaching... I'd love to know the details though if you could share more about that @VECTORBOY

geeksareforlife:
School meals are already free for all Key Stage 1 children (Foundation Stage 2 - Year 2), and personally, I think it is a great idea.

I would be a bit worried about funding though - at present, funding is decided on a "census" day - the number of children having meals that day is used to set the school's funding for an entire year.  This means that if you have a lot of absences that day, or even a school trip, funding will be lower than it needs to be (this happened to the school I am a Governor for!).

That is probably an implementation detail though, and in principle I fully agree.

Vote: ✅

Floppy:
I just updated this to resolve conflicts. Just wondering if the votebot will count the votes again... adding this comment to make it notice and recount.

Floppy:
Vote: ✅

Floppy:
@geeksareforlife can I ask you to recast your vote now the conflict with the master version has been resolved?

PR #511, opened 2017-04-06
National school uniform

philipjohn:
A single national uniform will be a lot cheaper, helping to deliver the real policy here - free school uniforms.


Autumn-Leah:
yup sure why not

Vote: ✅

Floppy:
OK, I see the advantage of free school uniforms. It is worth being more explicit about that being the aim?

Vote: ✅

philipjohn:
Yep! See #515

PR #513, opened 2017-04-13
Children and Families section

philipjohn:
There's a bunch of stuff I think we could do here that warrants it's own section.

### DIFF
--- a/children-and-families.md
+++ b/children-and-families.md
@@ -1,0 +1,4 @@
+---
+title: Children and Families
+---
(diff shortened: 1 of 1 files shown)


Floppy:
Yep - there's some stuff that could move in straight away.

Vote: ✅

Autumn-Leah:
Vote: ✅

Floppy:
I think this is good to merge, and it's also reminded me that @Autumn-Leah's 'equality' page, while created, never got linked into the menu, so I'll get that done at the same time.

PR #514, opened 2017-04-20
Majority needed for constitutional change

philipjohn:


Floppy:
Vote: ✅

andrewdwilliams:
The first word is "W" not "We". You missed the 'e'.

philipjohn:
Damn it! Thanks @andrewdwilliams - fixed it now.

Xyleneb:
How many of [these reforms](http://alevelpolitics.com/ukconstitutiontimeline/) would you have seen should they have been put to a referendum and held to this rule of majority?

Referendums are one of very few means to hold our representatives accountable, and what this essentially does is weaken their utility even further.

philipjohn:
Referendums don't hold politicians to account at all - they measure the will of the people in order to inform politicians. Holding politicians to account means scrutinising decisions they've already taken, not informing them which decision we'd like them to make.

As for past constitutional change, there's a whole bunch. Constitutional change includes things like major treaties, devolution, structure/makeup of Parliament. In particular, decisions like signing the Treaty of Rome, the Maastricht Treaty, and the Treaty of Lisbon, as well as the triggering of Article 50. I'd also suggest regional mayors, police and crime commissioners, and regional assemblies.

I could see future decisions about voting age, further devolution, trade treaties (like TTIP) all being subject to constitutional referendums under this policy.

Floppy:
Revoting because of typo fix. I think this does need to a tighter definition, but that's really hard given that the UK doesn't have a good definition of what constitutional change is. That needs to be worked out down the line, but I'm OK with the principle.

Vote: ✅

Autumn-Leah:
Remind me, since when was silence an opinion. If someone doesn't vote, it's because they don't care about voting and weren't going to and didn't, or they do care and were going to or wanted to but are unable.

In the first of these examples their opinion is irrelevant, since they don't have one, or at least aren't willing to provide it through the electoral system, they have definitely *not* said that they accept the status quo. A vote is an affirmation of an opinion and a desire for a representative or body of people to follow either their opinions and desires or to stand on the principles of of their own opinions and desires, not a passive acceptance.

In the second example, you're saying via this proposal that a government has the right to assert how a citizen would vote in any given election or referendum based on *their lack of participation in said election or referendum* despite the fact that have an opinion one way or the other that at that moment they were not able to express. This is in a very literal sense, rigging votes, and is fundamentally undemocratic.

As per the constitution of this party: https://github.com/SomethingNewUK/party-constitution/blob/master/party-constitution.md this is wrong. In said constitution it states in section 2.7:

> To influence the governing of the UK to perpetuate a nation that is a just, fair and tolerant and liberal democracy.

Now please explain what about asserting that a citizen voted a way they didn't through inaction, is in any way fair, just, tolerant of other political standpoints than the status quo, and in anyway suitable for a "liberal democracy"?

Additionally on the front page of the manifesto it states both:

> Bottom-up: People should be telling the government what they want from a local level, rather than being told from the top.

> We also believe that as a country, we can self-govern better in the future, and will create policy that hopes to re-engage citizens in our democratic process.

Now tell me how exactly this policy helps further engage the electorate into the process of representative democracy, as opposed to alienating them from it?

This is illiberal, immoral, unethical, and authoritarian, it also goes against the fundamental values of the party which you should know, since you helped make it, but of course you insist that Something New shouldn't exist and isn't a political party so you don't have to feel guilty when you attempt to push vote rigging onto the electorate and wider population.

This is being blocked. I'm disgusted.

Vote: 🚫

Polygon48k:
I think It should be something like 55%, doesn't parliament need a percentage to pass? would make sense to mirror that. We can't really say what non voters do, but in the manifesto it does include a reference to being able to vote non-of the above, That would be reasonable to interpret those votes in the manner described, a vote of no change or deffered decision.

Boundarybreaker:
@Autumn-Lean I want to ask you: How is abstaining from voting any different from voting to keep the status quo right now? That's what's been shown with Brexit in a massive way.

Floppy:
@Autumn-Leah - thanks for the input. I can see your point of view, but I personally don't think we're in the realms of vote-rigging here. 

There are many democratic thresholds in common use that require something other than a straight majority of votes cast. Needing [supermajorities](https://en.wikipedia.org/wiki/Supermajority) for organisational constitution changes is *really* common, as is the use of [Majority of the entire membership](https://en.wikipedia.org/wiki/Supermajority#Majority_of_the_entire_membership) (or absolute majority), which is what @philipjohn is proposing here.

The point is that we're taking major decisions on knife-edge votes; I don't think trying to improve that situation is a bad thing, personally.

Floppy:
Actually, I think the wording on "effectively recording an abstention as a vote for the status quo" isn't really correct, that's where the presumption comes in. Abstentions are abstentions, as @Autumn-Leah says, they aren't the same as status quo votes.

@philipjohn do you think we could drop that sentence? It wouldn't materially affect the idea of needing a majority of the membership.

Incidentally, this means guidance is needed on how such questions should be phrased. The brexit referendum question was "Should the United Kingdom remain a member of the European Union or leave the European Union?". If this suggestion was the law, presumably both answers would need a supermajority for the issue to be settled, otherwise it would be declared a draw?

The 1975 EEC referendum asked "Do you think that the United Kingdom should stay in the European Community (the Common Market)?". It was a 67% yes, but that wouldn't have passed this either, btw, with turnout meaning that only 43% of the electorate voted yes.

The AV referendum asked "Do you want the United Kingdom to adopt the 'alternative vote' system instead of the current 'first past the post' system for electing Members of Parliament to the House of Commons?". That's better phrased, and a supermajority requirement for change would make sense with the question.

philipjohn:
@Autumn-Leah 

> If someone doesn't vote, it's because they don't care about voting and weren't going to and didn't, or they do care and were going to or wanted to but are unable.

How do you know that? Do you have empirical evidence? If you don't, then we cannot possibly assume any reason for voter apathy.

> you're saying via this proposal that a government has the right to assert how a citizen would vote in any given election or referendum

That's absolutely not what this proposal does and, has @Floppy has mentioned, supermajorities and absolute majorities are very common in democratic systems. So to call this "illiberal, immoral, unethical, and authoritarian" is incredibly off base and the block on the grounds you've outlined is entirely baseless. I've put forward a proposal for using a well-known, well-used democratic system for votes constitutional change.

@Floppy Yep, that sentence is superfluous, so it can go.

The questions is a really interesting thing. There was a bunch of political wrangling over the EUref question. IIRC, these things are ultimately decided by the Electoral Commission but they obviously face a lot of lobbying, and I'm guessing approval by the government.

Floppy:
Vote: ✅

Floppy:
I think the removal of that second sentence reduces the implied assumption of how people would have voted, and now we're just talking about a fairly standard majority of members threshold for major constitution-changing action, which to me seems reasonable. @autumn-leah would you reconsider now that sentence is gone?

Xyleneb:
>The point is that we're taking major decisions on knife-edge votes; I don't think trying to improve that situation is a bad thing, personally.

There is already very little motivation for offering referendums to the public. Politicians don't like doing it and you're usually lucky to see one. So it's a rare opportunity for direct democracy. If they're already a weak hand that's seldom played, then I don't see how weakening them further is going to help them.

There's a couple of other points too. All of my votes are usually protest votes. Always the independent candidates. Always I vote for whatever the government doesn't want. I disagree with the notion that referendums are lacking the opportunity to hold those in civil employ to account. Every bit of space we're given is a Boaty McBoat opportunity, and unlike that particular incident these things are often treated as binding.

Then I can tell you about what I want:
I want 16 year olds to have the vote. With these referendum rules, might not happen.
I want to leave the EU. Ditto.
I want a mayor, not because I care _who_, but because for some reason he brings budgetary autonomy for my area. 30% of the electorate bothered to pick 'who'. The rest stayed home.
70% of the electorate showed up to pick an MP which is good, but the independents that I select get <1%.

Those who get to steer the boat are those who bother to vote. I don't want apathy to stall us until things finally become unbearable.

Autumn-Leah:
was about to change my vote, but @Xyleneb makes a very persuasive case. I think a good alternative would be to allow multi-day voting times rather than just a single day, to get more people to vote; then you'll get massive turn out because it's not an incovenience.

philipjohn:
> then you'll get massive turn out because it's not an incovenience.

What's your evidence to support that assumption? If people aren't bothering to vote, what makes you think they'll suddenly bother on a different day.

There are lots of things we could discuss for trying to increase *turnout* but this is about the legitimacy of the *result*, regardless of turnout, so anything to do with turnout is irrelevant to the proposal.

PR #515, opened 2017-04-20
Refine school uniform policy with reasoning.

philipjohn:


Floppy:
Vote: ✅

Floppy:
BOOM 💥

andrewdwilliams:
How do you envisage this national school uniform? Is it uniformity in every item of clothing? What about the inclusion of school badges, or differentiation of blazers or ties (very often different colour ties and blazers exist within schools, never mind between)? Isn't there a serious risk of a psychological effect, compounding the monotony of attending school? Also, different schools are different, and the competitiveness between those schools that is amplified by differing uniforms, creating a sense of identity, is healthy in that it promotes sporting and academic competition. I'm just floating ideas around but I can't imagine every school student wearing the same uniform in every school is going to work out very well.

philipjohn:
I could envisage the uniform having a school bags that's unique but yeah, for the most part, I think it should be identical.

A school uniform doesn't create competition - competitions create competition ;) I think a single national uniform could even help with a feeling of national unity.

Xyleneb:
I think you could really benefit from fleshing this out.

1. If you define the stitching and the material, but forbid schools from mandating the appearance then you would keep some of Leah's previous aim to retain freedom of expression. The pupils themselves would decide whether they want their school represented by red, or blue, or polkadot or zebra stripes. As long as the material and stitching is to spec then safety and flammability concerns shouldn't really come into it.

2. Can't a case be made for uniforms being free at the "point of use"?

3. If yes, then you must protect the uniform budget from both government and school budgetary decisions. You start rationing clothes, and you get the haves/have nots situation which you're trying to avoid.

4. Can the policy not be to facilitate a gentleman's agreement _between_ schools, rather than to command them with the law from above?

Get this done right and it'll afford the greatest liberty to the greatest number of people.

philipjohn:
1. That's fine detail that is far too much for the vision that the manifesto is supposed to be. The exact detail of implementation would end up being decided by ministers through consultation with educators and parents.
2. What does that mean?
3. Why would school uniforms be rationed? That's the opposite of this proposal!
4. I can't begin to express how much the phrase "gentleman's agreement" wants to make me throw my laptop out of the window... We make laws to confer unavoidable obligations where it is right and just. A "gentleman's agreement" obligates no-one and creates a free-for-all where only those with any sense of decency actually honour it.

> Get this done right and it'll afford the greatest liberty to the greatest number of people.

The greatest number of people is every single school child. Therefore, the policy at it stands benefits the greatest number of people it could! What more do you want?

VECTORBOY:
I would suggest a rqtionale would be eceonomy of scale, the basics range in many super markets has become the base of most uniforms now. Basically, it should be cheap and easily available.

Back in the day police used the uniform to round up scivers and take them back to the right school. 

Most schools (that I have visited) have dinner cards, logins or biometrics now though.

But i don't think an affordable uniform will stop schools having and promoting their own identity values and crests.

Autumn-Leah:
I've got to be honest with you, I don't know what I was thinking when I voted the original proposal through. This is a terrible idea. School uniforms are used on school trips to identify who's from what school, which will cause major safety concerns if all uniforms are standardised. You'd have to make the budget ridiculously large as quite a lot of kids lose or rip uniform and thus it needs to be replaced, and they also frequently grow out of it, this is without thinking about those experimenting with their gender identity and gender expression. Now whilst I fully support the rights and ability of everyone to openly express their gender and express their morphological freedoms as they see fit, having to constantly order and send out new uniforms to kids with no reference point as to how they'll express themselves is just going to drain money from local authorities that they simply don't have.

Now, what I *would* be in favour of is a national standardisation of uniform, so it's not provided by the state, but regulated by it so that kids don't have to wear unnecessary, expensive, or unnecessarily expensive clothing, or clothing that inhibits their education or gender identity, *and* I'd support a fund for trans people that can prove they're seeking medical treatment to get a free load of the "other sex's" uniform. But I can't support this proposal. There's no grounding in reality here, it's a matter of, "yeah sure lets just hand out uniforms that cost money and see if they get sold on ebay".

I'll now be making another proposal to get rid of the existing one.

Vote: ❎

philipjohn:
> School uniforms are used on school trips to identify who's from what school, which will cause major safety concerns if all uniforms are standardised.

I regularly take my kids on days out and end up bumping into numerous school groups. They are almost always wearing coats, hiding whatever school uniform is underneath, and quite often they are wearing hi-vis jackets on top of those coats. I've seen those hi-vis jackets come in lots of different colours, and often have the school logo on.

> Now, what I would be in favour of is a national standardisation of uniform

The uniform would just be the same then, and would suffer from the same problem of identification you've mentioned above.

> "yeah sure lets just hand out uniforms that cost money and see if they get sold on ebay"

If every school child gets free school uniform, who on earth is going to be buying them on eBay?!

Autumn-Leah:
@philipjohn Standardisation with allowance for colour differences and logos is different from handing everyone *the exact same uniform*. I'm taking about standardising materials, lengths, stitching, etc. so all kids get a decently made uniform, not handing them out.

Xyleneb:
Ok here are the points I've got:

1. I include details when I intend to clarify, or to enshrine (or cover the extent of) your rights. I'm not specifying "65:35 cotton-poly twill", because these are the fineries of the policy that do not enshrine your rights.

2. The aim of your policy is to make things cheaper for parents who cannot afford them. Did you not consider "Why stop there? Why not make them free?"

Yeah, midgets and the under-nourished might get free or next-to-free clothing if they really want white shirts and black trousers... but I'm not opposed to that happening. It could be considered a valuable public service and an investment in our young to clothe them.

 3. If schools were given control over the "free uniforms" supply, then they would tell you to take pride in it, constantly complain in your ears that you're treating it as expendable, and then try to ration the uniforms they give to you because replacing them costs money and their budgets are non-existent. I know it's the opposite aim. But that's what would happen if schools were custodians of the supply.

4. All of these changes you make ought to be done in the context of later handing these decisions over to the teachers themselves, in the form of a Ministry of Education. The role of Secretary of State for Education should not be making these decisions. I don't trust teachers to make the right decisions either so some means of public oversight should be included, but *anything* would be a step up from Michael Gove or a private company (e.g. AQA) deciding what you ought to learn or do in school.

With the notion of handing these decisions over in mind, you must give schools at least the illusion of choice, rather than ordering them from up on high... Or else they will immediately repeal most of what you aimed to do. Hence the suggestion for a gentleman's agreement, impotent as it may be.

>The greatest number of people is every single school child. Therefore, the policy at it stands benefits the greatest number of people it could! What more do you want?

Firstly, if you go too soviet on the idea of being utilitarian and of providing economic value then everyone will be left wearing olive-drab, at the expense of freedom of expression. This is not providing the greatest liberty to all. You could have made efforts to avoid it.

Secondly, our rules must pass the test of being in some way a necessity.
Nobody wants to see enforced a bunch of unnecessary and heavy-handed rules. Unless they're puritans. You make the case that living without these rules would cause undue hardship. Usually (more often than not) I would make the case that freedoms are worth suffering for even if the cost is great difficulty.

*Now then*. I know I've used a lot of words but I just have a couple more points to make:

Both the Netherlands and Finland have high-performing systems, despite one being results-driven and the other being holistic.

Both countries have a Ministry of Education that is separated of government.

Both countries have a policy of not requiring school uniforms:
https://archive.is/8DiHc
https://www.theguardian.com/commentisfree/2014/nov/07/wearing-school-uniform-doesnt-help-us-learn

This simplistic wisdom therefore, points to abolishing school uniforms. Much as I would like to not be bullied for wearing knock-off addidas, the system demonstrably works and the guardian article makes a good point that people find a way to pick on others if they really wanted.

If no one makes a policy of abolishing school uniforms, then at some point I will submit one.

philipjohn:
> Firstly, if you go too soviet on the idea of being utilitarian and of providing economic value then everyone will be left wearing olive-drab, at the expense of freedom of expression. This is not providing the greatest liberty to all. You could have made efforts to avoid it.

School children are already required to wear a school uniform, thus limiting their freedom of expression. This change just means their freedom of expression is limited with the same uniform rather than thousands of different ones.

> Both countries have a Ministry of Education that is separated of government.

I like this idea... could have sworn I read this in another manifesto recently ;)

> Both countries have a policy of not requiring school uniforms.... If no one makes a policy of abolishing school uniforms, then at some point I will submit one with the reasoning I've outlined.

Funny you say that, I'm actually much more inclined to want *no* uniforms at all. It'd get my vote, but I'll still hedge my bets with this one too!

philipjohn:
> Standardisation with allowance for colour differences and logos is different from handing everyone the exact same uniform. I'm taking about standardising materials, lengths, stitching, etc. so all kids get a decently made uniform, not handing them out.

1. Your earlier comment said nothing about standardising on "materials, lengths, stitching, etc"
2. Standardisation is exactly what this policy does - it makes the uniform standardised across all schools in the county. I don't understand your objection to this when you are arguing *for* standardisation yourself.
3. The standardisation is *separate* to providing free school uniforms. If you're not in favour of providing uniforms for free, fine, but that has nothing to do with the standardisation of them.

PR #517, opened 2017-04-24
End speaker democratic deficit

philipjohn:
Puts an end to the democratic deficit whereby constituents whose MP is the speaker are not adequately served and do not have the usual choice during elections.

### DIFF
--- a/elections.md
+++ b/elections.md
@@ -36,6 +36,8 @@ To deter time wasters, the current requirement of 10 nominations from the electo
 
 Prospective candidates must have a good connection with the constituency or ward they wish to represent. This enforces the connection between representatives and their constituents and helps to disqualify candidates that are "flown in" by the party ahead of local choices.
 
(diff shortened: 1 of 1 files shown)


Floppy:
Vote: ✅

davidmiller:
`they will be jobs advertised, filled and salaried just like any other.` according to received wisdom, most jobs are not advertised.

Even if that's not true, wouldn't this still be a weirdly passive aggressive tone to take in a manifesto?

philipjohn:
It probably is passive aggressive but that's because I'm angry ;) Personally, I'm not too bothered about that but I'm also open to changing the wording if you think it really needs it?

Autumn-Leah:
I'd change

>We will end the practice of the speaker's constituents being denied representation and a lack of choice in elections. The positions of speaker and deputy speaker will not be filled by MPs, they will be jobs advertised, filled and salaried just like any other.

to 

>We will end the practice of the speaker's constituents being denied representation and a lack of choice in elections. The positions of speaker and deputy speaker will not be filled by MPs, they will be jobs advertised, filled and salaried just like any other, as the alternative would be a continuous exception in the premise of this country's democracy.

Just to make it sound less angry and most justifiable, but I'm not stopping it.

Vote: ✅

PR #518, opened 2017-04-26
Remain in the EEA and EFTA post-brexit

Floppy:


philipjohn:
Vote: ✅

davidmiller:
@Floppy I don't think GH flavoured markdown supports quotes in markdown links. 

Also: IMHO airquoting stuff isn't helpful in this context :(

Also: If you link from `four freedoms` you should link to a page explaining the four freedoms no ? 

Also: If you're linking to the single market wikipedia page shouldn't the link be on `Single Market` ?

Floppy:
Thanks @davidmiller, I had a syntax error in the link. I should use the editor properly rather than hacking away by hand :)

Floppy:
The four freedoms links on Wikipedia do go to that Single Market page which explains them, but yeah, it was confusing, so I've moved it as suggested.

philipjohn:
Vote: ✅

Floppy:
I probably should have separated the "we will aim to stop Brexit" part into another proposal, but it is just a re-commitment to existing policy, so I don't *think* it should be a blocker here. I can do that though if it's a problem for acceptance.

andrewdwilliams:
Generally in favour of this idea. I think it's one I posited a while back too.

Vote: ✅

billiedreamwalker:
Hello James,
firstly, the article did bring me to SomethingElse.
I don't normally bother with Linux Format, unless it has an interesting article on Open Source, as this one did.
I preferred Linux Voice while it was a  Mag in it's own rightAnyway, I had been looking for somewhere to hang my hat for a while and I agree we seem to be on the same wave-lengh on a number of things.

I  didn't vote yet,because I am not clear what I am agreeing to.
Just support for remaining in the E.E.A. after Brexit or promoting the overturning of Brexit too?

I can support the former but not both.

I believe Brexit is the only direction for Britain, and  that unless Britain does leave the EU , there will be anything other than a promise of change in Europe that would not come.That is why I believe in Brexit so strongly.
It isn't just about Britain.I believe Europe needs Brexit as much as Britain does. 
At present the EU seem to believe they are calling a bluff, and that Brexit won't happen, so they are not seriously considering devolving more power to National Governments and electorates.

If ,or when , Britain leaves , then the reality will sink in that other Nations can do the same. Only then, do I believe that the EU will see  a need for substantial change.
I hope this helps to make my position a little clearer.

Sayonara!

Floppy:
This should be ready to merge now... further changes are always welcome.

PR #519, opened 2017-04-26
Strengthen commitment to basic income

Floppy:
Stated that we do want to provide a basic income in future. Kept the short-term policy of running research, and added a commitment to running pilot schemes.

### DIFF
--- a/economy.md
+++ b/economy.md
@@ -50,9 +50,9 @@ Close down all tax havens under British jurisdiction. This means reforming tax r
 
 In parallel, work at an international level to get rid of tax havens outside British jurisdiction, and close international loopholes such as the "Double Irish Dutch Sandwich".
 
-## Social Security
+## Basic Income
 
-We will support research into a [universal basic income](https://en.wikipedia.org/wiki/Basic_income) and its effects on society and the economy, with the aim of eradicating poverty and providing a basic standard of living for all.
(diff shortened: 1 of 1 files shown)


philipjohn:
Good improvement

Vote: ✅

andrewdwilliams:
Vote: ✅

Xyleneb:
I preferred the commitment to test projects, rather than the commitment to a national roll-out. I think the latter will make you unelectable (unless you consider the green party to be electable).

Floppy:
Polling is showing that this idea is becoming increasingly popular, so I don't think we're going too far in saying that we want to aim to get there. Passed, anyway, improvements welcome :)

LudovicD:
i am all for it but it is important to be properly and fully funded from economic rents, i.e.
Land Rental Value (& take >50% of Land rental Value over median level, & going to 62.5%, like above 2x median, & further/more as needed)
(& probably to fund it at least to 15% of GDP/head) (which i'd work out at £81-£84/week), & obviously that all legal residents get it.
never forget that basic income/unconditional benefits funded from non-Land/non-natural resource taxes, increase rents extracted from the recipients, so has very bad counter-productive effects.
But, as long as economic rents/Land rental values are well taxed, it is also good to tax income at the total household level, that is in excess of the median level of that household size, probably with a basic rate (as we have 20%), and then
for example in a mutiples
( 1.5x ; 2.25x ; 3.375x ; 5x ; 7x ; 9.6x ; 12.8x +) with higher bands (30%, 40%, 45%, 50%, 55%, 60%, 62.5%)
(these are total rates, only on incomes after all other non-income taxes (& after all other payments for public services or legal obligations or debt payments) constitutes income before tax for income tax).
(or 1.67x; 2.9x; 4.75x ; 8x ; 12.8x +) with (30%, 40%, 50%, 55%, 60%, 62.5%)

PR #521, opened 2017-04-29
Add Changing Places toilets in large public places

Floppy:
Because for people with serious disabilities, toilet floors are not an acceptable place to change. This is slightly tricky as the provision of toilets isn't a duty on any part of government, so I'm not exactly sure what the mechanism would be, but the intent is to set out the commitment to making it happen, with the detail of implementation to be worked out down the road.


philipjohn:
Planning laws might be one way to do it.

Vote: ✅

Autumn-Leah:
Can't see a downside to increased accessibility for disabled people to large events and city centres, aside from costing corporations more money if we mandate it, but stick em, they don't pay 90% of their bloody taxes anyway.

Vote: ✅

PR #523, opened 2017-05-06
Remove local areas

Floppy:
Now that we can easily run more projects with the same software, I think local manifestos are best done as separate projects, with a separate pool of contributors.


philipjohn:
Vote: ✅

Autumn-Leah:
Before I vote it in, I'd like to know how these separate manifestos are going to work, and why you think they're best done separately.

Autumn-Leah:
I'll have to vote no until then because until I know the details, local manifestos can be incredibly valuable to include on the same site.

Vote: ❎

philipjohn:
> I'd like to know how these separate manifestos are going to work

They'll be using the same system, so they will work in the same way. The actual vote numbers could be changed, and that may well make sense for a local manifesto that may have fewer contributors.

Floppy:
Hey @autumn-leah, thanks. The idea is that instead of having one single manifesto, we could set up individual manifestos for specifically local areas, as @philipjohn says. It's possible to set one up in about 10 minutes, these days. I've got one running for a proposed "North Horsham people's neighbourhood plan" at https://nhpp-votebot.herokuapp.com/, for instance.

Floppy:
In terms of why separate would be better, I think that contributions and conversations will be quite different in nature, and people who want to write a local one might not be the same as those who want to write a national one. But it's all experimental - all I know for sure is that the local pages in here never took off, and perhaps another more focused way might be better :)

geeksareforlife:
I think I agree with this in general, but would want to see the local projects linked from here in some way.

We'd also have to think about how it would work for parties (such as Something New) who are pointing at this manifesto, would there be official OpenPolitics local projects?

Vote: ✅

Floppy:
Yeah, I'd definitely agree we need a way of linking things together, but I think that happens outside the manifesto itself. I'd suggest that we add a list of active projects on https://openpolitics.org.uk, which could link to any projects using the software platform.

Floppy:
Tell you what, there are already a few projects running on the software stack, so I'll make a list of those, and then there will be somewhere for this content to go to. I'll get back to you.

PR #524, opened 2017-05-06
Generalise the equality section

Floppy:
As I suggested when this page was created, I think for now until we have a lot more there and a split is necessary, this page can cover equality in general, rather than just gender.

### DIFF
--- a/equality.md
+++ b/equality.md
@@ -1,0 +1,6 @@
+---
+title: Equality
+---
+## Egalitarianism
+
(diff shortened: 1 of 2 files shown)


philipjohn:
Vote: ✅

Autumn-Leah:
I was going to transfer a lot of stuff over but I'm very busy atm. I think the page should be kept.

Vote: ❎

Polygon48k:
Vote: ✅

Floppy:
Happy to do the moves if you can point to the bits you wanted to move @autumn-leah?

PR #528, opened 2017-05-13
Child Poverty

VECTORBOY:
I think we need a section for Child Poverty


Autumn-Leah:
On the basis that is makes no real contribution and it'll make the manifesto look a little ridiculous in the middle of an election, and take a week to clear up presuming someone makes an edit the day this gets in, I'm voting against.

Vote: ❎

Floppy:
Closing early as deprecated by #529.

PR #529, opened 2017-05-13
Child Poverty

Polygon48k:
A specific Child Poverty section is needed as many people have asked for it, I have included views from others and an exclusion that means that benefits cannot just be constantly raised when there is a clear and present danger for the child in the home.

### DIFF
--- a/poverty.md
+++ b/poverty.md
@@ -2,7 +2,19 @@
 title: Poverty
 ---
 
+
+
 We believe that until poverty is adequately addressed, inequality will continue to increase throughout the United Kingdom. Successive governments have failed to address the causes and some policies have served only to exacerbate the problem.
+
+## Child Poverty
+
+Reinstate the link between annual increases in benefit levels and inflation.
+
+Ensure that support with housing costs for families renting privately rises in line with increases in local rents.
+
+Establish a commission to consider how businesses can ensure that customers on a low income do not face paying the highest prices for goods and services
+
+A commitment to give children safe, state funded homes when the family is consistently demonstrated to be the source of poverty and or abuse.
 
 ## Low Pay and Job Security
 


philipjohn:
Good thinking on the new section! We should check whether there are similar/conflicting bits elsewhere in the manifesto as well - we might have addressed some of this stuff in some way elsewhere, so good to check.

Could you explain a little more about the intention behind this please?:

> demonstrated to be the source of poverty and or abuse.

The wording sounds a little funny to me. I *think* I know what you're getting at, but want to be sure :)

Polygon48k:
Yeah, you're right, perhaps malnutrition and/or physical/mental abuse.

Was inspired by an email from ecpc.org, I agree with their three commitments, I just wanted to include some kind of safeguarding against some of the situations I have encountered. You can give the parents all the money in the world but if it doesn't get to the children then it is of no use.

LudovicD:
Regarding housing costs support, it important to specify at the same time that this supports should be accounted in 2 parts:
part for Land & part for the building/Long-term maintenance costs of the Landlord, & the bulk (>50%, quickly going up to 62.5% above 2 times median level, and more according to the principle that taxing the return of economic rent does not affect any economic decision at all, all the way up to 100% of the Land rental value effectively received - adminstration costs of paying that tax) of the Land part (at least above the level of the median household of the Landlord) has to be taxed (to the the total LVT, national+local), whereas the part for the building/Long-term maintenance costs of the Landlord has to remain entirely untaxed, a must regardless of private bankers charging interests to the Landlord.
(obviously the priviledge of charging interest [as the money for the loan amount did not exist before the approbation of the loan] should be subordonate to as much as the interest needed being used to pay the LVT in porportion of the Land value & in proportion the capital not yet paid back by the borower, ie. as the money was created out of nothing else than the contract signature, to buy the land with the property, the interest on it includes the obligation of the part of LVT, in proportion of the part borrowed & not paid yet).
very important to specify that a high rate of LVT is a must, because if it not done, paying housing support is a huge subsidy the the landlord = the policy increases inequalities & poverty, which is totally inacceptable & we want to avoid that at ALL costs, as much as nuclear war & full collaspse of all civilisation (it is important to have strong words for this). it certainly is very unacceptable that any Landlord treat any economic rent (above median level), as ordinary income (it is a sure way of playing monopoly -> which only leads in the end to just the one person ends up owning 100% of all the UK) (even if what currently happens is very close to exactly what we want to avoid). So the principle of no subsidy to the rich is the more important one here.

Floppy:
Might need a reorganisation later to fit in with other things, but I'm in favour of the policies.

I *think* that the housing-benefit-rise-in-line-with-rent is a license to inflate prices as it stands, so we'll need to add some sort of rent control as well. Which I also think is a good idea, so I'm not objecting on that basis.

Vote: ✅

PR #530, opened 2017-05-14
National School Uniforms, my contentions and regrets

Autumn-Leah:
I've got to be honest with you, I don't know what I was thinking when I voted the original proposal through. This is a terrible idea. School uniforms are used on school trips to identify who's from what school, which will cause major safety concerns if all uniforms are identical. You'd have to make the budget ridiculously large as quite a lot of kids lose or rip uniform and thus it needs to be replaced, and they also frequently grow out of it, this is without thinking about those experimenting with their gender identity and gender expression. Now whilst I fully support the rights and ability of everyone to openly express their gender and express their morphological freedoms as they see fit, having to constantly order and send out new uniforms to kids with no reference point as to how they'll express themselves is just going to drain money from local authorities that they simply don't have.

Now, what I *would* be in favour of is a national standardisation of uniform (re: material, stitching, durability, etc.), so it's not provided by the state, but regulated by it so that kids don't have to wear unnecessary, expensive, or unnecessarily expensive clothing, or clothing that inhibits their education or gender identity, *and* I'd support a fund for trans people that can prove they're seeking medical treatment to get a free load of the "other sex's" uniform. But I can't support this proposal. There's no grounding in reality here, it's a matter of, "yeah sure lets just hand out uniforms that cost money and see if they get sold on ebay".

### DIFF
--- a/education.md
+++ b/education.md
@@ -10,10 +10,6 @@ Funding for Sure Start children's centres is to be restored, and any closed chil
 
 Education should be freely available to all to first degree level or equivalent. Tuition fees for university students should be abolished, as university education is of general benefit to society, and should be covered by general taxation.
 
-## School Uniform
-Individual school uniforms for every school will be replaced with a single national uniform, provided by the local education authority to every pupil at no cost.
-
-
 ## State funded, state run
 We will end academisation and return all state-funded schools to Local Education Authority (LEA) ownership and control. No school will be permitted to be state-funded without being under the ownership and control of the LEA.
 


Polygon48k:
Don't like it, some schools don't even have uniforms. In the case of trans I don't see any issue with a non-gendered uniform.
But the previous standardisation sounds acceptable, and also prevents a state monopoly issue

Autumn-Leah:
@Polygon48k

> Don't like it, some schools don't even have uniforms.

Isn't that rather thought terminating, patronising, and cliche? Like an abusive parent in a way, "if you don't like this food then you must not want food at all, you won't have any snacks when we get home". Unless I'm reading it wrong and you're saying that *you* don't like it, but then the sentence makes no sense.

> In the case of trans I don't see any issue with a non-gendered uniform.

I think you mean "trans people" or "transgender issues" there, not "trans". Language aside, there's no such thing as a non gendered uniform. Current cultural norms insist on the distinction between the masculine and feminine, and whether they are reinforced in school or not reinforced by forcing everyone to wear the same uniform there, both cis and trans people alike will continue to associate certain clothing with a certain sex or gender, and thus if you were to enforce a "gender neutral uniform" (which would require the banning of all other uniform otherwise gender divisions will rise once again depending on how many of what gender or sex pick what clothes) you'd make a sizeable portion of the school, both cis and trans, incredibly uncomfortable. Unless you're a gender abolitionist and you seek to get rid of gender roles, gender, and sexual organs as ideas from our society in general, in which case, good luck, but I'm not with you on that.

> But the previous standardisation sounds acceptable, and also prevents a state monopoly issue

The legistlation currently *in* the manifest forces a state monopoly on the production and provision of school uniforms, removing this policy removes that issue, I fail to see what you mean by this.

Floppy:
Thanks for reopening the debate, @autumn-leah, this is exactly how it should be done :)

I'll think more before I vote...

Polygon48k:
I was agreeing with the state not producing or funding the uniform. Sorry for the confusion.

Everyone I have met and dealt with has just said, "I'm trans". So I'm just using the language I see reflected in the community.

I think, it is very common for both women and men to wear trousers and a shirt or t-shirt. I don't think this robs them of any sort of gender identity, people will still do their hair, and pull bits up to mean something. People can demonstrate their masculinity and femininity in all manner of ways, not just the clothes they wear. But still, a lot of schools don't have uniforms, and it is common for schools to have trips where they all buy hoodies in the school colour's with a name on the back, that is what I have seen locally recently at least.

Polygon48k:
This has basically already happened by the market forces of super markets, schools have just adapted to fit the 'single black line' that is rolled out cheap. Does it need any regulation at all? Uniforms have come down massively in price since I was at school so I would basically just leave this alone

philipjohn:
I'll just paste my response from #515 :)

> School uniforms are used on school trips to identify who's from what school, which will cause major safety concerns if all uniforms are standardised.

I regularly take my kids on days out and end up bumping into numerous school groups. They are almost always wearing coats, hiding whatever school uniform is underneath, and quite often they are wearing hi-vis jackets on top of those coats. I've seen those hi-vis jackets come in lots of different colours, and often have the school logo on.

> Now, what I would be in favour of is a national standardisation of uniform

The uniform would just be the same then, and would suffer from the same problem of identification you've mentioned above.

> "yeah sure lets just hand out uniforms that cost money and see if they get sold on ebay"

If every school child gets free school uniform, who on earth is going to be buying them on eBay?!

PR #531, opened 2017-05-14
NS&I

Polygon48k:
Seems like this kind of already exists, but is under utilised.

### DIFF
--- a/economy.md
+++ b/economy.md
@@ -4,15 +4,15 @@ title: Economy
 
 Markets are a tool for society, not as the primary reason for it. The economy should work for everyone, not just billionaires, hedge funds and investment banks. Excessive [rent-seeking](https://en.wikipedia.org/wiki/Rent-seeking) (gaining wealth from from financial speculation and intellectual property) increases inequality. In the short to medium term, we seek to rebalance the economy to reduce this by ensuring wage income is closer to rentier income.
 
-## National Investment Bank
+## National Savings & Investments
 
-Re-nationalise the UK Green Investment Bank and rename and restructure it as the National Investment Bank (NIB). The NIB will also be merged with the Public Works Loan Board and the Infrastructure and Projects Authority.
-
-The NIB shall be capitalised with £20 billion, drawn primarily from the sale of the government’s shares of private banks following the financial crisis, the [removal of subsidies for fossil fuels](http://openpolitics.org.uk/manifesto/environment.html#subsidies), and the existing funds held by the UK Green Investment Bank.
+National Savings & Investments shall be capitalised from the sale of the government’s shares of private banks following the financial crisis, the [removal of subsidies for fossil fuels](http://openpolitics.org.uk/manifesto/environment.html#subsidies), and the existing funds held by the UK Green Investment Bank.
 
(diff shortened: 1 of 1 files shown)


Autumn-Leah:
you might want to cite a source in the policy as to what NSI actually *is* in more detail because I doubt many will know even given what you've provided.

Polygon48k:
In principle could they cover the same goals?

Would this be enough?

https://www.gov.uk/government/organisations/ns-i

LudovicD:
No, the resulting bank (whatever it is called) should have capital (like £20 billion), & should be operated as a public service.
The right of private banks to create money should be taken away from them (if needed split them between a non-money creation part & a nationalised part), be back in an accountable committee of the BoE, who should create sovereign money in all forms (electronic & other), in quantity with its mandate of reducing inequalities, ensuring acces to income from activity/work, & with minimum for those who are currently denied a job due to lack of opportunities cristallising, & limiting inflation (while favorising actual wealth of meeting needs over physical growth of use of finite resources) (obviously, they will need to be directed by people who can think out-the-box, as it requires new systems to replace the old systems they have used before). ALL fuels should not only not be subsidised at all, but be taxed at 100% of extracted value (~ international market value - [costs of extraction & normal remuneration profit [4-5% of capital costs of extration]] ) at source for greenhouse gas producing fossil fuel [nuclear fission should be replacing carbon as a stop gap, only until nuclear fusion replaces it, which will happen as soon as possible, and a lot sooner than the 23rd century anyway] [the scalabiliy of renewable energy is quite low: it has the potential to reduce the share of non-renewable energy but (if including nuclear fission in non-renewables, only) by about 0.48% by year / ~ 12.8% every generation][considering the growth of population & of consumption of all other non-energy resources], so the transition is coming in front of us over the next 7 to 9 generations (during which nuclear energy is required as a temporary stop gap), obviously not taxing the greenhouse gas at source will slow down the process further (a further slow down is the least issue here, as, long before the 23rd century, castastrophique global warming will have runaway (which only a small minority of the human population can survive the scamble it would cause) if fossile fuels are not replaced 98% by nuclear fission [first a s a p, then by fusion when ready]), it certainly is already far too late to try scale up only renewables, as the technologies to scale up these renewable as fast as it would be neccesary do not actually exist yet : the (renewable) technologies that currently exists (or can be developped fast enough) are technologies that cannot be scaled up fast enough to avoid runaway global warming without scaling nuclear fissions up to replace 96 to 98% of the other fossile fuels, a s a p
(we certainly cannot afford to waste another 5-10 years election cycle) [ the need is not 1 Hincley B : it is to bring up (double up) capacity to about ~ 30 to 40 nuclear power plants for each country the size of the UK, or about as much as France had at its maximum nuclear].

Floppy:
I *think* they're different (though I'll defer to an expert). NS&I is a citizen savings bank, which allows citizens to invest in government bonds etc, to encourage saving? The NIB would be an agency making infrastructure-type investments.

I'm going to have to abstain as I just don't know enough about the area.

Vote: 🤐

philipjohn:
Yep, NS&I is really more like a high street bank and does a very different job than an investment bank.

PR #532, opened 2017-05-14
Multiple Changes

Polygon48k:
Fiscal studies are already part of the curriculum. encourage learners to learn when they are ready, and encourage those that have completed studies to not take advantage of the state. removed free uniform. Should probably of split this.

### DIFF
--- a/education.md
+++ b/education.md
@@ -10,8 +10,12 @@ Funding for Sure Start children's centres is to be restored, and any closed chil
 
 Education should be freely available to all to first degree level or equivalent. Tuition fees for university students should be abolished, as university education is of general benefit to society, and should be covered by general taxation.
 
+We will encourage young people to apply to University when they are ready, rather than committing to an expensive, life changing choice due to convention and expectation.
+
+We will fund this by removing state funding for Further Education for plateaued students. 
+
 ## School Uniform
-Individual school uniforms for every school will be replaced with a single national uniform, provided by the local education authority to every pupil at no cost.
+Individual school uniforms for every school will be replaced with a single national uniform.
 
 
 ## State funded, state run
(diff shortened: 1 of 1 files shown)


Autumn-Leah:
Definite should have split this. Removing funding for uni is a terrible idea, and if it's not as bad as it sounds then it definitely needs clarification before I can vote it through. Free uniform is a policy proposal I've already created, and thus it'd be redundant to have another one, if you want it changed, go vote my change through since we seem to agree. Fiscal studies isn't actually part of the core curriculum yet, and if you can point me to the bit of legislation on gov.uk that says that I'll be delighted because it means that the government did something vaguely correct and justified for once.

Vote: ❎

Polygon48k:
Yeah, its a bit all over, I support free Higher Education(Uni), but not pressure into it. And a cap on the Further Education system (College). i  was looking at https://www.pfeg.org/about-us/about-pfeg but after some investigating, although being delivered in many schools now, it is not on the National Curriculum. Perhaps we could agree on commiting to the PFEG's goals.

Floppy:
There's some stuff here that would go straight in, and some which will need more discussion. So that we can discuss each in turn, can you please resubmit them as separate edits @Polygon48k?

philipjohn:
Removing funding for FE and the free school uniform are both a big no from me in any case, I'm afraid.

Vote: ❎

LudovicD:
yes to both,
all general education should be from general taxation. (Further education should be with grant on the understanding of exercising in the UK for a full decade, post-grad, is actually mostly doctors, research/engineering, lawers anyway, which all bring value to the country) (& research/development UK budgets need to be multplied anyway)
for uniform, they should allow small variation(s) per county council/city council.

PR #533, opened 2017-05-14
Violent Crime

Polygon48k:
A proposal to remove criminal elements from our streets, having an impact on the exposure of the next generation to this behaviour.

### DIFF
--- a/crime.md
+++ b/crime.md
@@ -2,6 +2,10 @@
 title: "Crime &amp; Justice"
 ---
 
+## Violent Crime
+
+In every corner of the country we see repeat offenders going in and out of prison, receiving little punishment and becoming emboldened by their knowledge of the system. We propose a three strike action for violent criminals who have caused serious risk to life on three occasions. They would receive life imprisonment. Currently, crime is under reported due to a very small number of Violent Criminals who are well known to residents and Police, which undermines the values of the majority of the population. We will support the right to defend yourself from the results of Violent Crime,
+
 ## Capital Punishment
 
 We are opposed to the death penalty for all crimes, and would vote against any attempt to reintroduce it. We will also lobby internationally to stop the death penalty in other countries.


Autumn-Leah:
Yeah, this isn't a great idea either, though it depends on your perspective on the function of prisons. If you see them as punitive then when taken to extremes, the medieval torture chambers of yesteryear can be applauded. If you see them as reformatory institutions like I do, then you'd understand that 1) by putting more people into prisons you're merely increasing their chances of repeat offence due to being surrounded by the very people who are better than them at doing whatever it was that got them in jail in the first place 2) jails often don't act as reformation centres and are thus poorly equipped to allow newly freed people back into society, meaning that they come back out with less skills and experience than they should have, and then due to the criminal record can't enter the workforce. They then promptly begin to starve and turn back to crime as their only option. Jails thus create a self fulfilling prophecy of criminality both when jailed and when let out. If jail reform were to be effective you'd be treating the inmates like people by giving them decent facilities, teaching them skills they need to re-enter the workforce, and allowing them to vote (which we currently don't allow: https://en.wikipedia.org/wiki/Felony_disenfranchisement#United_Kingdom). But of course then we'd have people complaining that they've got decent facilities in prison and that that's unfair, because that's apparently how fairness works. : /

I recommend you read the UN's guide called 'Alternatives to Imprisonment': https://www.unodc.org/pdf/criminal_justice/07-80478_ebook.pdf

Vote: ❎

Polygon48k:
I have some understanding of alternate systems, but, the primary concern should be their removal from society to do no additional harm. What and how to deal with them should be secondary to this. I know Prisons are full and underfunded but, along with the decriminalisation of drugs from the other page it may level out. In regard to fairness it is simply a statement of fact that newly built prison cells are larger than the student halls rooms I attended.

I would suggest that some, not all, will never be 'rehabilitated' and should therefor not suffer cruel conditions, but neither those preferable to the 'working man'. They should be able to be productive and useful to society, whilst being kept separate.

As for voting, we should certainly allow political prisoners to vote (if we do in fact hold any). But I am not for normal prisoners voting. There should be removal of privileges, but not the introduction of any particular punishment, other than the attempt to mirror the daily working routine outside of prisons.

The UN guide doesn't seem to be working so I will have a look for an alternate source. To my knowledge, these have only been successful in countries with "checkered" past's concerning the treatment of the vulnerable, ie. Nordic countries.

Polygon48k:
https://www.unodc.org/documents/justice-and-prison-reform/cjat_eng/3_Alternatives_Incarceration.pdf is this the same document?

philipjohn:
There are already a [number of offences that carry an automatic life sentence after the *second* offence](http://www.cps.gov.uk/legal/l_to_o/mandatory_and_minimum_custodial_sentences/#automatic). I'm not sure if this would add anything new?

Floppy:
^ what @philipjohn said.

Also, what does "We will support the right to defend yourself from the results of Violent Crime" mean, and how is it different from the current state? (I am not an expert in this area in any way, so forgive my questions)

Autumn-Leah:
@Polygon48k > As for voting, we should certainly allow political prisoners to vote (if we do in fact hold any). But I am not for normal prisoners voting. There should be removal of privileges, but not the introduction of any particular punishment, other than the attempt to mirror the daily working routine outside of prisons.

Voting isn't a privilege, it's a right, and I support all prisoners having that right.

PR #534, opened 2017-05-14
Nuclear Disarmerment

Polygon48k:
Sensible steps towards reducing worldwide stock, rather than throwing away our immediate defence programs. Britain has reduced their arsenal greatly, that should be highllighted.

### DIFF
--- a/military.md
+++ b/military.md
@@ -4,9 +4,7 @@ title: Military
 
 ## Nuclear Disarmament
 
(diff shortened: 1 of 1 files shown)


Autumn-Leah:
> The UK should encourage European and NATO partners to share nuclear resources as a way of reducing the total number of war heads in existence.

Seems like a better idea than current policy if you could just clarify this for me.

Polygon48k:
I remember watching a video of French general proposing a reasonable way to reduce warheads without reducing near term capability. I can't find the source though...

Floppy:
Interesting idea. Could we keep some of the original wording, and use the "pooling" as a stepping stone to full disarmament? We've lost the future commitment part of the original policy, which (from my reading) you'd still be in favour of?

PR #535, opened 2017-05-14
Adjustment to arms policy

Autumn-Leah:


Polygon48k:
should be "The United Kingdom of Great Britain and Northern Ireland", I have never seen that acronym and it yields no google results so I would drop and, as it is only used twice use the full name of The United Kingdom of Great Britain and Northern Ireland.

Polygon48k:
:negative_squared_cross_mark:

Floppy:
Yeah, that's an odd acronym :)

I'd support a change to UK-built though, rather than British.

Floppy:
Forgot to vote, reasoning above :)

Vote: ❎

philipjohn:
What they ^ said :)

Vote: ❎

PR #549, opened 2017-05-21
updated the living wage figure (outside London) (added to no tax on tampons, other hygiene products)

LudovicD:
- updated the living wage figure (from outside London, living wage fondation) (also added mention of London figure, hour & year)
- added to no tax on tampons, other basic hygiene products (in the interest of hygiene in the whole society)

### DIFF
--- a/economy.md
+++ b/economy.md
@@ -16,7 +16,7 @@ The NIB will target particularly deprived or underfunded regions, and local auth
 
 ## Minimum Wage
 
-The minimum wage should be set to the same value as the living wage (currently £14,458.50 based on 252 working days of 7.5 hours per day at £7.65), and automatically adjusted in line with that figure.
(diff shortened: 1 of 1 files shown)


Floppy:
Thanks for updating the figures! There's nothing here I disagree with, but could I ask you to separate it into separate proposals? There's a couple of bits to discuss in each, that I think would be worth doing separately.

LudovicD:
How do i separate it now ? ( with this git hub, proposal, which is new to me )
(1. the figures, 
2 the addition (of other hygiene products) to no tax on tampons
3 the ownership (man-made) )
( i think the definition for me of the commons include all natural resources, so Land, that are not human-made, which is why the logical way of specifying ownership as private is to be explicit with qualifying the something to be bought-sold [->transfered to private ownership] as human-made ).

Floppy:
Best way is to go back to the page you edited and do a new suggestion, one for each separate part. Then we can close this one. Does that make sense?

LudovicD:
yes, a bit the long way, but i will do that.

LudovicD:
done, split, so when the 3 are seen available for vote, we could close this one.

Floppy:
Thanks, this is now split into #560 #559 and #558. Closing this one early :)

PR #550, opened 2017-05-21
specification in buisness tax (vs taxes on household) (revnue vs profit)

LudovicD:
more specification in buisness tax (vs taxes on household) (revnue vs profit)
with the ideas to avoid double taxation (anything that is a service vital to the population count as tax/payment for public service, even it is/became private).


Floppy:
I think this is generally OK, but it's very hard to read. We might be better off separating the bracketed sections into footnotes, for further explanation without making the text unclear?

Floppy:
Looking back at the original text, that's already quite hard to read, so it's not just your additions @LudovicD :)

Autumn-Leah:
I don't want to vote this through because I can't decipher it. Care to make it more readable?

Vote: ❎

Autumn-Leah:
I know that's a silly reason to vote against, but if I can't be bothered, neither can the average voter.

LudovicD:
I changed it,
to be written much more clearly.
It is basically to ask for a study, and specify the actual parameters of the study:
(= the difference between the gross all-encompasing figure of turn-over that includes everything that is taken away from us & the net figure that is actually what is possible to play with or use [by households or by company]).

Regards.
Ludovic

Xyleneb:
Discounting the wording problems which you definitely have got, I like the initial premise but I don't know how you'd budget it.

Let's say you make £37,500 a year and that the first £17,500 is tax-free, while the remaining £20,000 is flat rated at 25%. You would pay £5,000 into the government's pocket.

Then let's say you make £37,500 a year but only managed to put £2,000 away into your savings account. At the same 25% flat rate, you'd be paying £500 into the government's pocket.

The changes you propose (under this example) would hit the government's yearly income from income tax by a factor of 10. How would you still manage funding for government services?

LudovicD:
all public expenses comes by theory out of economic rent (longue introduction),so it is natural to propose funding governement services, by levies(fees) on values extracted from economic rent,mainly economic rent of natural resources(mainly rental value of unimproved Land):private tenants have already this charge included in their existing rent, so it only need to be charged on theLandowners (Landlords &amp; owner-occupiers), as a super-council tax (&amp;super-business rates) only on the rental value of the Land, (because it is charged on benficiaries, owner-occuppier with a mortgage,can deduce the economic rent represented by the bank extra profit in the form of interest higher than admin costs)(the bank would be then charged some levies on the economic rent they receive, as the definition of a mortgage is the bank owns the property &amp; Land, in proportion of the loan remaining, and let the user use it while they pay back),in a way that home-owner (at the start of the loan) would not pay more in total (mortgage+Land levies/taxes) than if they were tenants (&amp; the banks would be charged the levies/tax part not charged to mortgagees/home-owners).(this, added with a tax on carbon &amp; one on pollution)replacing council tax, business rates, stamp duty, insurance tax, road tax, TV Licence, and VAT &amp; some income taxbelow old lower threshold/NIC.(for the interest of making these fees/levies on rental value of Land only (e.g. median house £240k, Land: £130k building: £110k, rent: total:£12k/yr, Land:£6.5k/yr, building:£5.5k/yr)(Land is a smaller proportion at lower value), progressive, the rate would be low below median [level of Land value ownership amongst all households, including those who do own any], and above median high(>50%), and continously increasing, to discourage greed in extraction, the remainder after fee/levy would be asympototicly capped as in a serie of inverse of roots of degrees 3 [large Land market size][quicker increase] to 6 or 9[small flater Land market size][slower increase]).
But the required taxes on profit are not just a tax rate on profit put away, the profit is calculated byincome minus the normal needs for different expenses categories (each capped, most at something like 110 or 120% of median or average for that household size) (&amp; small amounts for the leisure category). so if there is £17.5k tax-free for that houshold size, it probaly be up to £22k-£25k for the cap of everyhing (above: all is profit).so £35.5k would be for a larger household size (like 2adults+3kids) instead of (2adults+1kids for the amount above)so £500 charge for them, but the state would save a lot in not paying them any cash benefits (such as child benefit, winter fuel) (it would have paid in other systems as the tax-free &amp; the maximum cap are for net zero cash benefit/zero contribution point).The principle is that most people are charged all kind of various rents (line rental, for TV, for compulsary insurance on car, road, tax, MOT, etc., for water, fix per day for energy,,.., public transport pass per unit of time, tax on hospital/car park, tax on teeth/dental, tax on eyes/optics), and they end up with huge whole household costs of living, more than earning normally minus income tax(&amp;NICs), and have no control of being thrown into poverty. So they should have cover for the unavoidable costs, and get taxed only on profits they get above the various (mostly rental) costs.the actual tax should (1) start from zero, (2) stay low below median level (household),(3) when getting to 1.5x time income, 2x, 2.5x, 3x 4x, 5x 6x, 8x etc... tax rate getting from [variable0to10%], [variable10to20%] [variable20to30%],  [variable30to40%], [variable40to50%], [variable50to60%], 60%, 62.5%[HMRC estimate for the point where higher rates lower the tax received anyway], so it is seen to progressive tax,not a flat rate, but for the super-rich, a high flat rate is ok
Regards.Ludovic DELANNOY



----Original Message----

philipjohn:
I'm not keen on this, not because I don't agree with the aim, but because it's too detailed. For me it veers to far into the specifics of what the study would actually look into. Whether or not things like utilities, insurance etc are included need to be considered separately by the study. All we should do in the manifesto is say we'll kick off the study, and let the study (and therefore the evidence) give us the specific answers.

Vote: ❎

Floppy:
Fair point there @philipjohn. This is becoming more of a regular thing as people flesh out existing policies; we should come up with some sort of guidance on it, or a way of handling extra detail. I've opened a ticket for general discussion on that in https://github.com/openpolitics/manifesto/issues/578.

Personally I'd be OK for now with this level of detail in a footnote or two.

PR #551, opened 2017-05-22
Added new section about House Building

areteh:
The biggest driver of house prices is the restriction on supply. These proposals plan to increase the availability but making the construction and modification of individual units faster, cheaper and easier.

### DIFF
--- a/housing.md
+++ b/housing.md
@@ -41,8 +41,10 @@ If a privately run housing association seeks to sell properties because they are
 We would encourage the creation of housing co-operatives by those who own their own apartment(s)/flat(s)/house(s). Housing co-operatives allow the value of direct democracy to (literally) hit home, allow tenants to escape landlord exploitation, and result in the maximization of available living space.
 
 ## House Building
-Invert the current planning regime i.e. planning is automatically granted unless the Council can demonstrate within 8 weeks that the proposal contravenes policies (local or national) and/or building regulations.
+Update the current planning regime such that:
+a) planning is automatically granted unless the Council can demonstrate within 8 weeks that the proposal contravenes policies (local or national) and/or building regulations
+b) any changes to local planning policies would require full public consultation. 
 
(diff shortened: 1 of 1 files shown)


Floppy:
In the first paragraph, is the intention that planning should be automatically granted? The "invert the current regime, i.e." part at the beginning is confusing me, as I'm not sure which bit is the policy and what is the current state. I'd suggest removing that to turn it into a simple statement of policy.

I agree that restrictions on supply are a problem, especially at the affordable housing level, but I'm concerned at the reduction of residents' right to object to things that "don't fit" in their local area. That gets people hugely concerned. I'd rather see encouragement to build in ways that would get local residents on side, rather than ignoring them. Does that make sense? I think I'm a no for now, until we can phrase this in a way that doesn't trample on people's wishes and put developers in the driving seat.

Vote: ❎

Floppy:
A bit of context, sorry: this has been a big issue near me recently. People don't feel that their wishes are taken into account, and are very unhappy with it. I'd be wary of making that situation worse.

areteh:
Understood. I have duly updated the wording to make the first point clear. The original idea came from the existing planning regime in Germany. The idea is to make it easier for the good stuff to get through i.e. the stuff that always would/should have but too often gets tied up. It was never intended to give free reign to the bad stuff - developments or developers. To this end, I added the aspect of greater public involvement in setting those local policies in the first place.

The second paragraph is largely driven by issues around brick construction. It has terrible insulating properties resulting in higher CO2 emissions and making pensioners poorer. It is also a painstakingly slow building process and the inaccuracies prevent standardisation of other parts. The problem is that planning is very hard to achieve on non-brick construction because of the "in keeping" clause.

Floppy:
Oh, interesting, thankyou! I will have a think :)

Autumn-Leah:
Vote: ✅

PR #552, opened 2017-05-22
Added sub section to improve competition and market operation

areteh:
The initial privatisation of the railways did not result in free and fair competition. The single national monopoly (British Rail) was simply replaced by a series of private companies effectively operating monopolies on their routes. The only competition is when these operators bid for the rights to operate their monopoly. In contrast some other European countries introduced competition by allowing private operators to work in parallel with the national system. For example in Austria the passenger has the choice between multiple operators on the same route offering different prices, quality and speed. 

This proposal therefore seeks to have free and fair competition at the level the customer experiences and not just the level the Department of Transport experiences. Moreover, it would be done in a way that enables local communities to serve their own unique needs.

### DIFF
--- a/transport.md
+++ b/transport.md
@@ -72,6 +72,9 @@ Investigate feasibility of building a high speed rail network that connects our 
 
 Examine which areas of the United Kingdom were most detrimentally affected by the changes in the [Beeching Reports](https://en.wikipedia.org/wiki/Beeching_cuts), and whether it would be economically feasible to bring back any formerly established railways.
 
+Introduce true/genuine competition by allowing multiple operators on the same routes and replace the series of monopolies that currently exists. Operating licenses could be granted to companies operating at the level of single train (or even carriage) so that communities  can act to better serve local needs. 
+
+
 ## Bus
 
 Promote efficient use of road space in towns and cities by allocating road space to bus lanes, encouraging the creation of a network of frequent bus routes.


Floppy:
I was thinking that this is in conflict with a statement to nationalise the railways, but we don't actually have that, so yes, I'm in favour of a better model for competition. The current system is a total hack.

As a side note, we should perhaps flesh out something on the renationalisation (or community ownership in some way) of the rail infrastructure (lines, stations, etc) if not the services that run on it - a lot like roads. Do you know how that's done in other countries?

Vote: ✅

geeksareforlife:
I like this idea - we could flesh it out a bit more looking at how subsidies are allocated etc, but a good starting point.

Vote: ✅

philipjohn:
My feeling is that we should just get rid of profit (which the manifesto already does) and competition entirely and go straight to nationalisation. There is clear public support for it, and we have the evidence to show that it returns money to the treasury.

Vote: ❎

areteh:
This idea was based on the Austrian model - to which I added that it could go down to level of single trains to help local communities. There the track system is state owned and there is also a state owned operator. The private operators increase customer choice by varying price, frequency, speed and quality of carriages. So there is competition both between the private operators but also with the national one too. This drives efficiency savings in the national operator and makes them more responsive to passenger needs. 

[On a side note China still has many state owned companies but interestingly has saught to introduce competition here too. For example in telecoms, there are no private operators (for reasons of state security) but they broke up the state monopoly into a couple of state owned companies that are in direct competition with one another. Again this has resulted in both cost savings and increased customer responsiveness)

Floppy:
I like that approach @areteh, especially of allowing both public and private operators. I think we've got a chance here to propose something different, that isn't straight 1970s nationalisation, or broken markets, but something better. Something new :)

(this is now me thinking out loud)

I'm really interested in models of public ownership that aren't government ownership. Seems to me that's something that nobody's talking about right now, and could fit alongside competition to drive efficiency.

At the moment we have the worst of all worlds. We have public infrastructure that's not run for the benefit of the public, and also has no competition so prices go up and services get worse. If we had a system where rolling stock operators were still in competition with each other, but were owned publicly or at least had a public interest CIC-type clause in their constitutions, perhaps that would give the benefits we seek from a hybrid approach. I think this is similar to our NHS thoughts.

All that's a long way of saying @philipjohn that I don't think that competition and public ownership are totally incompatible. But we need to work out what that forward direction is. I think this proposal is an interesting step towards it though.

Vote: ✅

Autumn-Leah:
Introducing capitalism to the individual train level doesn't minimise it's inefficiencies.

@Floppy

> I'm really interested in models of public ownership that aren't government ownership. 

How about collective ownership by the people that run the train and the stations it stops at.

Vote: ❎

areteh:
That would certainly improve pay and conditions for the workers but what about the customers? Without competition how could we ensure better services and more affordable prices? Isn't this why there was a move away from collective ownership in the first place (albeit at the national level)?

PR #554, opened 2017-05-22
Use of crypto and local currencies  

areteh:
The 2008 financial crisis had devastating impact because of how inter-linked the global financial system has become. Since then much regulation has been introduced in the attempt to increase the safety of individual institutions and the national system. However, nothing has been done to break-up the links globally. These proposals attempt to create a parallel system that would then be able to continue functioning in the event of any future crises.

### DIFF
--- a/economy.md
+++ b/economy.md
@@ -94,6 +94,12 @@ All public sector organisations will have to use open source software and servic
 
 We will instruct HM Treasury to produce a roadmap for the future of currency and the role of central banks in a world with non-state-issued digital currencies, such as Bitcoin.
 
+Actively facilitate wider use of Bitcoin and other cryptocurrencies to create greater heterogeneity in monetary supply and therefore greater resilience to future financial crises.
+
+Encourage development of local/regional currencies to reduce food miles, stimulate local economies, create greater transparency between producer and consumer; and encourage greater sense of community.
+
+Potentially facilitate mechanisms to link the aforementioned cryptocurrencies and local currencies so that economic gains in one local currency can readily be used in  another area too. Thereby creating significant parts of the economy that run in parallel but importantly would be isolated from contagion of any toxic financial products created in the mainstream banking sector.
+
 ## Land Ownership
 
 Follow recommendations regarding transparency in the [Lyons Housing Review](http://www.yourbritain.org.uk/uploads/editor/files/The_Lyons_Housing_Review_2.pdf) - "To ensure greater transparency in the land market, the Land Registry should open up land ownership


Floppy:
I'm totally unconvinced on the utility of local currencies, especially when they are joined together into larger areas (isn't that just reinventing the Pound?), but I know that many people cleverer than me are in favour. And I definitely think we need innovation in the area of improving our finance system, so I think I'm a yes, especially as this is framed as "encourage" and "explore".

Vote: ✅

geeksareforlife:
I would really love to see some references - why are local currencies good?  I fully accept that I probably know nothing here, but that makes me feel like I can't vote either way!

Autumn-Leah:
Frankly I just don't know enough about the relationship between regional and national economics in this area.

Vote: 🤐

areteh:
There are several potential advantages to local currencies but for me the main ones are the environmental one (through the shortening of local supply chains) and that more of the value created stays in the local community (rather than ending up in the City/Wall Street). The first is reasonably easy to verify but as yet there is less empirical evidence for the second - it is admittedly a more theoretical arguement at the moment. In keeping with the desire for "evidence based policy" in this manifesto, I have framed this as "encourage/explore". The idea being that with more local currencies in play a bigger data set becomes available - and one that may also allow for more apples to apples analysis.

The linking aspect is a bit more complicated. The disadvantage of a local currency is that it only allows for a certain specialisation of producers - and there could therefore be an upper limit on how sophisticated a local economy could become. For more specialised goods/services a larger market is needed than the local area may be able to provide. In Catalonia, they have overcome this by linking 30+ cooperatives using multiple local currencies through a cryptocurrency. See http://cooperativa.cat/en/ and https://fair-coin.org/

That such a linked system would duplicate some of the function of the Pound is indeed true. However, this is partly purposeful in order to build greater resilience against future financial crises. One of the reasons that the banks were bailed out on such favourable terms in 2008 was that we were only days away from the total breakdown of the whole financial system i.e. the ATMs and electronic payments would have stopped. There have been insufficient reforms to ensure that the same could not happen agai. However, by having a parallel system, we could potentially have large sections of the economy that could continue to function even in such circumstance - meaning that a Government could not be held to ransom so readily.

geeksareforlife:
I have been talking to someone with more knowledge than me in this area and it seems like it could be a really good idea.

His point though is that you really have to be all in (but you can move to the full system gradually) - he is suggesting that you would have multiple local "pounds" that are all linked to the national GBP - that is traded on international markets. All transactions take place electronically and therefore you can easily have people buying things from region B when they are in region A - stimulating growth in regions.

It seems that the ultimate end to that would be large amounts of money coming from London to the North, as they can buy their bread/milk/cheese/televisions cheaper by harnessing their "strong" pound. This implies to me that the environmental advantages would end up being wiped out, but could have really good economic advantages.

I'm still not sure about it, but thought I would share!

philipjohn:
I'm also unconvinced about the local currencies, and I'm also unconvinced that cryptocurrencies would have had any impact on the financial crisis. In fact, just the wild fluctuations in Bitcoin make me question that.

It'd be good to get some sources/evidence about how local currencies help stimulate the local economy.

Vote: 🤐

Autumn-Leah:
This seems like a bloody good idea :D

Vote: ✅

areteh:
Thanks @Autumn-Leah 

I realise now I could have framed the original proposal better. Aside from creating greater heterogeneity into the financial system it is also fundamentally an issue of monetary supply and thefore monetary policy. Ultimately it is about how do we want money to be created - who by and who for. 

At present >90% of money is created by the priivate banking sector - primarily as new debt. This is what creates the debt bubbles and ultimately makes the "real economy" subservient to the private banks. This issue is well addressed by the Poistive Money group - http://positivemoney.org/our-proposals/

In sumary it calls for 
1. Take the power to create money away from the banks, and return it to a democratic transparent and accountable process
2. Create money free of debt
3. Put new money into the real economy rather than financial markets and property bubbles

Unfortunately, I am not very optimistic about this level of reform happening at the level of national government - but more so at a local level. The direct human contact can aid the accountability and transparency required for democratic governance.

Moreover, it is notable that a single currency and interest rate has seen different parts of the UK fare very differently over the last 5-10 years. By putting monetary supply into local control it would allow local inflation to be controlled according to local needs.

I see the linking of these local currencies as a necessary step to overcome the limitation of critical mass for specialisation. Unfortunately, direct evidence is limited here as neither local nor cryptocurrencies are exactly widespread. The Catalunia example is the best I have found thusfar.

PR #555, opened 2017-05-22
Savings in Governement Procurement and using it to stimulate innovation (in private SMEs)

areteh:
Multiple governments in other countries have benefited from transitioning to open source software. This proposal is also i line with our own desires to use open source wherever possible - whether software or manifestos! 

It also suggests a mechanism by which Government procurement (not just in software) can stimulate innovation and the success of start-ups.

### DIFF
--- a/economy.md
+++ b/economy.md
@@ -84,6 +84,15 @@ All private companies who receive public money must be subject to the same trans
 
 [^secret-contracts]: *"Stop Secret Contracts"*: [http://stopsecretcontracts.org/](http://stopsecretcontracts.org/)
 
+Drive savings by transitioning all Government departments (local and national) to open source software as and when licenses are due for renewal. Use these savings to provide every child with a Raspberry Pi.
+
+Stimulate innovation on the demand side by requiring that 5% of Government Procurement comes from companies meeting 2 of the following 3 criteria:
+	* Fewer than 20 employees
+	* Established within the last 3 years
+	* A spin-out of UK University technology
+
+
+
 ## Public Sector
 
 In order to stimulate adoption of the living wage, and act as a good example, all public sector organisations and [National Infrastructure Organisations](infrastructure.html) will be required to pay employees at least a living wage.
--- a/economy.md
+++ b/economy.md
@@ -84,8 +84,6 @@ All private companies who receive public money must be subject to the same trans
 
 [^secret-contracts]: *"Stop Secret Contracts"*: [http://stopsecretcontracts.org/](http://stopsecretcontracts.org/)
 
-Drive savings by transitioning all Government departments (local and national) to open source software as and when licenses are due for renewal. Use these savings to provide every child with a Raspberry Pi.
-
 Stimulate innovation on the demand side by requiring that 5% of Government Procurement comes from companies meeting 2 of the following 3 criteria:
 	* Fewer than 20 employees
 	* Established within the last 3 years


Floppy:
The first half of the first paragraph is, I think, already covered in the next section.

> "All public sector organisations will have to use open source software and services where available. Proprietary software that is currently in use should be examined for open source alternatives. A timeline should be created and made public to show how and when the body will convert to this alternative in a reasonable period of time. This is to cut down on public sector costs and increase our society’s participation in the open source culture."

But I like the idea of the rest, so I'll vote yes on that. 

Can I suggest removing the first sentence, and moving the raspberry pi statement to a separate proposal in the education section? I think there's more that can be said there, and we can always say where the funding would come from in that section.

Vote: ✅

geeksareforlife:
Whilst I love open source, we have to recognise that it is unfortunately not always the right choice, for a variety of reasons (some of which we can have a positive impact on):

First and foremost, is the skills to use that software. We can obviously provide training on this, but it ranges from the high-volume low-skill training for office staff, to low-volume high-skill training for people like graphic designers.

The second issue is an application's suitability. You will get no argument from me when it comes to server technologies - open source leads the way here - but many desktop applications are not equivalent. Most office users can get away with LibreOffice, but many will need the features that they get with Microsoft Office (The document foundation actually doa  good job of documenting this: [basic list](https://wiki.documentfoundation.org/Development/Really_basic_missing_features_and_enhancements) [big table!](https://wiki.documentfoundation.org/Feature_Comparison:_LibreOffice_-_Microsoft_Office).
When we get into the realms of more specialised software like Adobe products, eLearning software, Analytics tools, then the gulf becomes even wider.

The third, and possibly easiest to solve, issue is expectations from outside of government - people expect a word document when they have to fill out a form, etc.  This can fairly easily be overcome, but will have an initial startup cost associated with it that will probably wipe out any first year savings.

As for Raspberry Pi - yes, I love them! I own two. But I'm not sure that this in keeping with the principle of the idea.  The Pi runs open source software, yes, but it is not an open platform, nor is it open source hardware. Surely an Arduino Yun would be more appropriate? I am sort of joking - I would love to see money going to get children more involved in maker/hackspace culture, but I'm not sure buying a raspberry pi for each of them is the way to go.

I like the procurement bit though.  If we can remove the open source bit (especially as it is covered in the next paragraph) then I will happily vote yes.

Floppy:
Vote: ✅

Autumn-Leah:
Vote: ✅

geeksareforlife:
Vote: ✅

Xyleneb:
I'm not sure what penalties there are for failing to hit it. Which would make it simply some small commitment to the electorate that wasn't met...

Listing the consequences (whatever they are) for failing to hit the target would give this submission substance.

From what I'm able to understand from these things, is that contracts don't happen with a lot of vendors because the contract writing process is complicated by it. So they end up as monopolies that use a single company to attain 100% capacity (usually g4s, BT, balfour beatty, etc). Or in the case of the olympics: "not quite 100% after all".

It is the contract writing process and it's complicated nature that is the problem. Small enterprises are competitive naturally as long as you don't ask them to meet 100% of the work. They haven't got the size or the staff to do it.

With that in mind the best I could do is vote zippy :zipper_mouth_face:

I like how the Arduino is mentioned in submission #555...

PR #557, opened 2017-05-22
Decentralisation of taxation

areteh:
At present on ~5% of taxation is in the hands of local authorities. This creates a situation where central government tax rates and spending may benefit one local community but benefit another. In contrast is Switzerland, where the population of 4 million are spread across 26 cantons - each with the power to set rates and spend according to local needs (and only a small fraction being collected as a "federal tax").  This proposal supports a move in that direction for the benefit of both thriving and deprived parts of the UK.

### DIFF
--- a/economy.md
+++ b/economy.md
@@ -26,6 +26,8 @@ Tax laws and the tax system must be simplified. Much tax avoidance happens becau
 
 As a general rule of thumb, tax policy should focus on simple direct taxation (such as income tax, corporation tax and land value tax) for the bulk of revenue raising, and use minimal indirect taxes to discourage certain activities (such as fossil fuel use, drug use, and so on).
 
+Enable greater decentralisation of taxation by enabling local authorities to set tax rates according to local needs, keep those revenues and spend them according to local need. 
+
 ### Income Tax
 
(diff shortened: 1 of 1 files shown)


Floppy:
Yep, this is in line with a similar statement in https://openpolitics.org.uk/manifesto/democracy.html#devolution on federal and local taxation.

Vote: ✅

Autumn-Leah:
Vote: ✅

RobertWootton:
I think that perhaps to qualify for a personal tax allowance, a person must be a member of their local Residents Association that is registered with the local authority. If there was not an already existing R.A. a person must start an association. A person who lived "Offshore" would not be eligible for any tax allowances.

areteh:
Please note that this proposal is not about an allowance but rather who sets the actual tax rate in the first place. It is about decentralising those tax rate decisions

RobertWootton:
My personal view is that the existing taxes paid by an electorate should be paid directly to the local parish/town council and a percentage of that amount paid to the district council and a percentage of their receipts paid to the County Council and a percentage of their receipts paid to Central Government of the UK. Who decides on the percentages is open to debate.

PR #558, opened 2017-05-22
updated the living wage figure (outside London) (+mention of London figures hour & year)

LudovicD:


Floppy:
Yep - helpful to update the numbers. Might be worth editing to say "as of May 2017" rather than "currently". We could also move the bracketed stuff into a footnote to keep it easy to read.

Vote: ✅

Autumn-Leah:
Vote: ✅

PR #560, opened 2017-05-22
(human-made) added to ownership

LudovicD:
(human-made) added to ownership, as to distinguish property suceptible to be bought from the commons/public property which it is illegitimate to take away from the public owner & so public property can only be rented (never privatised, such as the sea, the water, space, wild life, the countryside, the road network) (all Land ownership has to be compensated by some form of LVT (even with an allowance, or with a relatively low starting rate) wherever the exclusive use of that Land is in private hands [no Land if its rental value is non-null should be exempted: even if the Land belong to a public entity with exclusive use: if it belong to the state/the UK, as crown dependencies, LVT is due to be paid to the council, if it belongs to the council, LVT is to be paid to the UK state]), not bought, and has to include all natural resources, including Land

### DIFF
--- a/economy.md
+++ b/economy.md
@@ -101,4 +101,4 @@ information to the public in a similar manner as the property price paid data se
 requirement to register land option agreements, transactions and prices."
 
 ## Ownership
-When you buy something, it should be owned (not a license to use). A preventative measure to reduce risk of vendor lock-in and promote ability to fix and modify purchased products.
+When you buy something (human-made), it should be owned (not a license to use). A preventative measure to reduce risk of vendor lock-in and promote ability to fix and modify purchased products.


Floppy:
OK, yes, to exclude natural resources, I get it. I think it might need a little more explanation to be clear in the manifesto itself, but I agree with the principle.

Vote: ✅

Autumn-Leah:
Vote: ✅

Autumn-Leah:
Actually wait, clarify it, or you'll get 3d printed things as exceptions.

Vote: ❎

LudovicD:
I think it clear that 3D-printed things is human-made & NOT natural ressource materials.
the working definition of natural resources materials is BEFORE they are extracted & are legitimately sold [with an eventual required tax] / shipped to their next user, to be transformed or 3-D printed.
please re-consider the vote.

Xyleneb:
I agree it's clear enough that "robot-made things" is not an exception to the rule. You wouldn't call them "natural objects".

What I disagree with is the principle that I can only rent the land on which my house is built; rather than own it outright as long as I am personally using it. Indeed I can claim to own the sea. But only the part of it that I am wading in or fishing in. If you say I can't own the sea, does that mean I can be bullied out of my wading spot because the government claims ownership? or that I owe rent? or democratic mandate? or something else?

Frankly I could live without this clarification; that injects ideology on the means of ownership into what was otherwise a point about copyright and digital rights management.

Vote: No

LudovicD:
that is the problem, it is all the citizens together claiming common ownership(because humans are biological with needs of nature) (not the state claiming, the state is obvioulsy only the referee acting on everyone's behalf), everyone living in the same country, everyone needs to have some space,so some sharing (of value) is needed for Land, but for wading spots, the sea is big enough for all.the value of the natural resources is built up by everyone.what is unfair is those who don't have and are not compensated because some have taken/used more of it without payingfor a fair part of the value that is created by the rest of society.rights &amp; copyrights are to be about things that are human-made,not about excluding some people from natural resources(the value of) (because these people as all, actually ultimately need some share of these resources)(which should not be denied without sharing the value of it)the principle is that the state is there to do good to everybody,but most of the bad &amp; wrong that has been done (allegdely by states) to some people,has been done by denying ('bulied out of') resources, opportunities, meeting of needs ... to these people.each other's freedom is always going to be limited by each other's freedom;everyone's freedom relies on everyone else's freedom; same for rights, the state is there to be fair to all,to protect some space for everybody, hence those who have not been allocated somewhere, by having been denied the opportunity of some Land for ownership, deserve compensation for a share of the value of their share of space they don't have, so they deserve a share of usage fees to charge to those using a share of the natural resource that is the space they own.it is only because of human needs of nature, that some deserve a share in value because they don't have a share in resources, and because that share in value should be charged to those who use more than their share of resources from nature, by choosing to own more (&amp; in an ownership way of excluding everyone else from it).this is the theory (of biological needs) anyway, but it should be the same in practice.
You see with levy on production: people produces because they have needs. a share above some level needs is levied to provided for those who are denied (by whatever reason) the opportunity to produce, so all can have something to meet some level of needs ; Levy or usage fees on natural resources are exacly the same, except they are not on anything produced, but they are on what has always been there to be shared, and are on things either extracted or used in an ownership way that exclude everyone else/the others from it.The principles are theorised for individuals, but actual living space in practice is inhabited by household(s), which why it is logical that copyright (&amp;digital rights)(&amp; property rights) apply for things that are created from(/by) an individual (&amp; usually only one individual at time enjoy the physical copyrighted things, because some people don't need anything of that category/of that kind), but copyright does not apply to things that are obviously from no-one, or from all(so their property rights are of a different kind)(=the kind of a relative share), and are needed by unit of household anyway, but everyone need some (space).I think it goes better with some clarification.Regards.Ludovic DELANNOY



----Original Message----

Xyleneb:
You make it seem as if natural objects are 100% excluded, 100% of the time.
That's my issue with it.

I can't own the water in my bottle.
I can't own the oil in my bucket.
I can't own the shiny rocks in my jewellery.

Somehow I must rent these things... Seriously?? Will I get a guy sapping my electricity because; "you can't, like, own it maaaan"?

This clarification does more to confuse things than clear things up. I don't think you should mix the policy about copyright and terms of use screwing people over, with the policy on nationalised industry.

We can argue over the ideological definition and purpose for property if you want, but the two points above are why I'd leave my vote opposed.

LudovicD:
It looks to me that it is clearly written that the natural substances are only excluded from private prperty before they are extracted (in small/fair amount or paid for) (&amp; not after/not 100% of the time) (in a bottle, bucket, container, or  rocks before they get packaged to go to a jeweller workshop to be cut &amp; become shiny).rent is only there as the usage fee to correspond to the value that everyone/everyone else in the society has contributed to,&amp; everyone should get their fair share from what has clearly be made from all.Electricity is always transformed (where it is not lightning), so no longer a natural resource (although there is a strong case that the main Large generation of it should be publicly owned, before it is supplied/distributed to consumer(s) as their share, or some of it sold to them).I don't think the case is good as to leave ownership without private ownership denied to commons that should never be taken from the commons.everybody knows that every time the earth was taken away from the bankers, the bankers have just always created the money, and bought it back again (especially since 1694). I have not seen much wrong that has not come from injustice, mostly from economic injustice.Limitation of copyright &amp; of ownership is at the center of the foundation for the civilisation to stand on.Regards.Ludovic DELANNOY



----Original Message----

PR #561, opened 2017-05-23
Move disability from being the last entry to earlier

petrolhead276:
Disability is getting a rough time in the UK and needs a more promonate position.
In percentage terms it is significantly larger than the LGBT percentage but with a lesser voice, this needs to change as anyone can become disabled or be born disabled and they are the forgotten "majority" minority group.

### DIFF
--- a/index.md
+++ b/index.md
@@ -14,7 +14,7 @@ We believe that government should be:
 
 We believe that individuals should:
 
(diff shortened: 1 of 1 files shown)


Floppy:
That's fair. Perhaps we should revisit this whole thing with that in mind and try to organise by population?

Vote: ✅

Autumn-Leah:
Vote: ✅

PR #566, opened 2017-05-26
Move toilet seat orientation from being no entry to earlier

Xyleneb:
Joking aside, what I want you to understand is the egalitarian point of view. That to create lists of people is to discriminate between them; to select your favourites, and then to leave others out. The commitment to treat people fairly shouldn't end at this one category or another. Everyone we've over-looked counts for something. Even the bald men, they count too! Furthermore, it is very hard to oppose a party that offers freedom for all, without any exceptions. It's very easy to oppose one that offers to single me out (whether or not they aim to be favourable or unfavourable).

Perhaps I'm the only one who sees it, and perhaps I'm as unconvincing here as I was over the contributor guidelines earlier on. But this problem will get worse as policy after policy tries to be specific to some amongst others, until they become so large and incomprehensible, or until enough people see it in the same way.

Freedom for all - that is the point.

### DIFF
--- a/index.md
+++ b/index.md
@@ -14,7 +14,7 @@ We believe that government should be:
 
 We believe that individuals should:
 
-* Be treated equally by the law, regardless of gender identity, sex, romantic orientation, sexual orientation, ethnicity, race, perceived ethnicity and/or race, age, religious and/or spiritual belief, occupation (current or previous), status of union membership (including the type of union and what union), marital status, or disability.
+* Be treated equally by the law, regardless of gender identity, sex, romantic orientation, sexual orientation, ethnicity, race, perceived ethnicity and/or race, age, religious and/or spiritual belief, occupation (current or previous), status of union membership (including the type of union and what union), marital status, beer consumption, home game attendance, nose hair length, quietness, loudness, lack of interest in handbags, computer game high-score, using the last bacon rasher, slouching, gym or no gym membership, not caring about some scene out of the lion king, putting ones foot in it, snoring, toilet seat orientation, belief, or disability.
 * Be free to act in any manner that does not harm another individual, and does not infringe upon the rights of other individuals.
 * Enjoy a fundamental right to privacy from the state or their agents.
 


Floppy:
I get your point, and it's true, but I don't see any harm in noting areas that have been, and currently are, a huge issue for the communities affected. We need to redress some of that, and acknowledgement is part of it.

This is obviously not intended as a serious edit, though the discussion is perfectly valid. I'll pop a block on it to prevent merging, but we can continue discussion, naturally.

Vote: 🚫

Floppy:
We could perhaps say "Be treated equally by the law, regardless of factors including but in no way limited to". But it's getting to be a real mouthful of a paragraph already...

Xyleneb:
>We could perhaps say "Be treated equally by the law, regardless of factors including but in no way limited to". But it's getting to be a real mouthful of a paragraph already...

If you think about it, the commitment to treat people equally and fairly under the law (regardless of this or that) is not much of a commitment. Everyone will claim it. Unless you detail how you're going to do it.

It's not just the empty sort of wording that I have a problem with though. My concern is that substantive edits as well will become bogged down by these discussions over specific assurances.

You might have a homelessness epidemic. You might have a joblessness epidemic. Wiltshire wants this. Cumbria wants that (I think I was thinking about pork sausages). They all require different levels of assistance and urgency. But all of them, _every last one of them_ must be treated the same. You cannot deny it to others and still consider yourself fair to them.

I might just be rambling to myself again, but until you reach that conclusion it's going to be a lot of time and energy spent discussing who makes the cut where and who doesn't. You'll never settle who deserves top place at the oppression olympics, because it's nigh on impossible to gauge these things fairly.

>This is obviously not intended as a serious edit, though the discussion is perfectly valid. I'll pop a block on it to prevent merging, but we can continue discussion, naturally.

As a test case, I don't think it's necessary to have a rule blocking joke entries.

Discounting any rule on the seriousness of submissions, what would you do if we democratically decided to all have free owls and to have all historic monuments renamed to 'Boaty Mcboat'? Would you fork the manifesto to something different? Would you change Something New's commitment to following it to: "with the exceptions of (x), (y) and (z)"?

Floppy:
No, I wouldn't want to make a rule automatically blocking anything - that's what the public votes are for.

Honestly, if someone did propose something that was blocked as being "silly", but enough people wanted it and could convince a block voter to remove that block, so in it goes? I think it's all part of the process. But the system is designed to be a bit defensive against stuff like that, as it's harder to get stuff in than reject it.

As for the actual section under discussion; the general vagueness of it is why this bit is not actually policy - it's a statement of belief and principle, not about any specific change to the law.

Floppy:
It's about stating something about the world we'd like to see :)

Xyleneb:
>Honestly, if someone did propose something that was blocked as being "silly", but enough people wanted it and could convince a block voter to remove that block, so in it goes?

That's the reasoning of an absolute monarchy. "Perhaps if enough of the subjects shout about it...". A case can be made that there should be no block-option at all, even if the submissions get all Hitlery.

>It's about stating something about the world we'd like to see :)

Ok, well, even the BNP will claim to treat people fairly under the law, but then their policies include rounding up all of the foreign looking people onto a boat and then sink the boat. I'm an expert at wasting words (half of which I made up) while barely putting a point together. Fact is I don't need idle reassurances from a politician. What policies you got?

More to the point it's not just about the empty stuff. I submit stuff on transport policy, and then I can't get phil onboard in part because I don't give enough reassurances to truckers and logistics businesses. Now, I love truckers. I've got friends and family in this line of work. And it's true that the policy I propose affects them the most. But I really don't want to target them with the wording or with any particularity, whether favourably or not. Because to do so would be to have bias. In view of the same friends and family.

I say "more to the point" I'm reiterating the point from earlier about not choosing favourites.

The fact is that these discussions over who should come where in a list will come up again, and they will get very long and very heated. You can't spend time productively arguing who should come first in a queue between a homeless orphan and a double-amputee. When they come up again I hope somebody will take note of the rhetoric I've given here. I think the only way to settle such a dispute, when the time comes, will be a consensus vote to simply delete the lists.

I could be wrong. But if this experiment grows then you could end up approving things like the progressive stack: https://en.wikipedia.org/wiki/Progressive_stack in good faith, only to find that you're being told to shut up based on your melanin pigment, and then to find everyone abandoning it like they did with the occupy movement when extreme partisans secured a foothold there.

For as long as you attract a majority of left-wingers who _mean_ well, you're gonna be vulnerable to stuff like this.

Floppy:
See #580 for my attempt to address some of this.

PR #559, opened 2017-05-22
added basic hygiene products (toilet paper, etc) to no VAT on tampons

LudovicD:
(other basic hygiene products too) (thinking of toilet paper, soap, shampoo, toothbrush, toothpaste & why not deodorant?)

PR #567, opened 2017-05-26
Replacing VAT with a Luxury Goods Tax

Xyleneb:
Not only will the government not charge you for basic hygiene - it also will not charge you for basic living standards.

Things will be cheap, cheap, cheap!
Other things will be dear, dear, dear... (covering the shortfalls of VAT will be hard). If you want a cheap watch, buy it 2nd-hand. Get it? 2nd hand? Why does nobody pay me for my jokes?

### DIFF
--- a/economy.md
+++ b/economy.md
@@ -34,11 +34,9 @@ Capital Gains Tax will be replaced by an Asset Tax, based on an annual self-asse
 
 The rate of Asset Tax for individuals will be harmonised with the rate of tax on earned income, and all allowances for each will be pooled, in order to ensure that work is rewarded and incentivised to the same extent as investment.
 
-### VAT
-Investigate possible reforms to the way VAT is applied to goods and services with a view to making it less regressive and, ultimately, to remove the poorest from having to pay entirely. This may require renegotiating our obligations to the EU, which makes VAT compulsory for all member states.
+### Replacement of VAT with a Luxury Goods Tax
 
-#### Remove Tampon Tax
-Tampons should not be considered a luxury item and should be considered a necessity 
+All goods categorised as non-critical, non-staple and non-utilitarian will incur a luxury goods tax. All other goods are to receive exemptions. For example: beer, chocolate, furniture, basic data retrieval and internet services, etc will be exempt from Luxury Goods Tax. However for the purposes of this tax, modes of transportation shall be categorised as a luxury good.
 
 ### Financial Transaction Tax
 
--- a/economy.md
+++ b/economy.md
@@ -36,7 +36,7 @@ The rate of Asset Tax for individuals will be harmonised with the rate of tax on
 
 ### Replacement of VAT with a Luxury Goods Tax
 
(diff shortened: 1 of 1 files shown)


Xyleneb:
Ohh and sorry I one-upped proposal #559 it is what got me remembering this in the first place.

Floppy:
Sounds sensible. I'm not sure how this would affect businesses, we might need some other category for business services perhaps.

Vote: ✅

geeksareforlife:
I like the idea, but why would you exempt beer and chocolate from the tax, surely they are luxuries?

Just trying to understand the rationale, I think the general idea is definite win.

Xyleneb:
The rationale is very hard to settle upon, because what is "not a necessity" is infinitely debatable.

You can't not eat. You can't not wear clothes (unless you're fine with hibernating come winter). But some food is £3,000 per head. Some clothes are too. What is "not necessary" essentially becomes reduced to things like clothing accessories, cars, entertainment and foreign holidays.

You could argue that these are necessities too, and arguably you could be right.

With all of that said though, the VAT system and it's exemptions are equally arbitrary. So in that sense I'm proposing to replace like with like.

Defining things in greater detail for example whether "educational software" should trump "entertainment magazines" is something which I would aim to leave out of the basic tenet of this proposal. I think it would be right to eventually cost everything and to explain everything, just in supporting documents rather than in the manifesto.

The aim of it is to reduce the regressive nature of VAT. Truth be told, I have doubts about whether it will work. You can afford to live more cheaply. So you can afford more luxury. But the cost of luxuries has gone up since the luxury goods tax. Have things then, really changed?

I think we can improve upon VAT, but whether it will be as revolutionary as you might want it to be is another story.

>Sounds sensible. I'm not sure how this would affect businesses, we might need some other category for business services perhaps.

I can't tell what you're thinking, unless you mean their VAT-exempt status.
The premise that the last buyer in the chain pays the tax could easily be carried over to 'LGT'.

geeksareforlife:
That makes complete sense to me, and having supporting documents is definitely the way to go!

Maybe we want to leave all examples out of the manifesto then, with just the tampon tax called out as it is a hot issue at the moment?

philipjohn:
I had the exact same thought as @geeksforlife - I like the idea but let's leave the examples out. The actual specifics of what is and is not a luxury item is something that will have to be subject to the same sort of system that decides what good are subject to VAT, so not something we can realistically set out in the manifesto.

Vote: 🤐

Xyleneb:
I think that the electorate would regularly ask for a few examples, and that the ones I selected would be of particular interest to the majority of public and businesses. It's one of those "define what rights you get" situations again.

That's why I want to keep it the way it is.

Apart from my one "No" vote, I can't stop it from changing after the fact in some edit or other.

Floppy:
I'm entirely happy that we can edit for readability later, and if necessary move examples into a footnote, as I've done with some other recent additions. The principle is sound.

Vote: ✅

Floppy:
What's really interesting about this is that (I think, I Am Not An Economist) give an incentive to reduce prices, to make things "not luxury" so as to avoid the tax.

I there would definitely need to be some predictable formula for working out what is and isn't affected (maybe % markup?), but that's beyond the scope of this.

Definitely reforming VAT into something non-regressive is a great goal.

Vote: ✅

Autumn-Leah:
Vote: ✅

PR #569, opened 2017-05-28
Limited time in office

Motobiman:


Floppy:
Does that mean that the two terms would have to be non-consecutive?

Motobiman:
No Sir

Two terms consecutively or not and then not able stand for ten years.

AndyC

Floppy:
OK - I agree with the principle though we might need to make the wording a bit clearer on that aspect. Perhaps "no more than two consecutive terms, and after that may not stand for re-election within two terms." (suggest change from 10 years because we might lose the fixed-term parliaments act soon)

Vote: ✅

Floppy:
To pre-empt objections of "but you'll lose all their expertise", most expertise on issues of policy comes from the civil service, researchers, and advisors anyway. We might lose some expertise at playing politics, but I don't think that's a bad thing.

Motobiman:
Sounds good.

philipjohn:
I don't agree with arbitrary limits on all serving politicians. At the moment at least politicians are self-selecting, which means that - locally especially - we would very quickly run out of people willing to stand for election and we'd have a constitutional chaos.

But more than that, this would have the effect of banning those representatives who do a very good job for their constituents, and are repeatedly elected because of that.

Yes there is a problem of "career politicians" but this proposal solves that in a very detrimental way.

Vote: ❎

Floppy:
Dunno @philipjohn, I think there are always enough people willing to do the job - we're never short of people standing for election, after all, and there's always competition to be selected at a party level. I agree that there are some MPs who do a great job at a local level, but I think the perception that it's a "job for life" once you're in in most seats is very bad for the system.

A quick look on Wikipedia (https://en.wikipedia.org/wiki/Term_limit and https://en.wikipedia.org/wiki/List_of_political_term_limits) doesn't show anywhere that does this at the representative level; anyone know of anything already out there?

Motobiman:
Hello Phillip.

Good point except with respect you are putting the cart before the horse and ignoring the big picture.

Government is there to run the country on the principle of what is good for the majority is good for all.

There is no doubt that Politics in Britain has developed into a business with party at its core.

That business employs people that accept they are going to be controlled by the party system no matter if it conflicts with their own ideas or those of their constituents.

Your elected representative on a local level may be effective in some small way but is entirely neutered by the parliamentary system that relies upon party support at every stage.

Those representative answer to and are different controlled their local party at every meeting and their only recourse in the event of disagreement is resignation therefore parliament is controlled by unelected party committees meeting in secret and exercising control over your representative.

In regard to MP numbers, that is another issue altogether, but with a properly constituted system in place and remembering we already have devolved governments in Scotland Wales and NI (but not England) and an extensive and ever growing local and regional political system, GB probably only not needs 100 or so MPs anyway.

The first step in dealing with career politics and the party control of elected representatives is time limited service.

This is by no means the only step we need to take to re align government to apply the democratic principle of the needs of the many over the wants of the few but it is a necessary step.

AndyC

Motobiman:
And people voting before the debate is another issue but without an answer.

philipjohn:
> I think there are always enough people willing to do the job - we're never short of people standing for election, after all

That's perhaps true at some levels, but definitely not all. The more local you get, the quicker we run out of people. Take a typical parish or district ward of a few thousand people - we already get co-option a lot (three times in the last 6 months near me alone) because of a lack of candidates, and limiting representatives to two terms will only exacerbate that.

For offices such as mayors, prime ministers or presidents a term limit could be a good way to prevent those offices being held onto by the same person if the controlling party is able to keep them there despite public dissatisfaction. Term limits still address the symptom though, and not the cause, which is ineffective mechanisms for recall or removing an elected representative.

We should instead make sure that elected representatives are not able to cling on to power, or change the rules in their favour whilst in power (see Putin/Morsi) rather than arbitrary limits that may end service of a good representative who has clear public support.

Motobiman:
Hiya

Comments below?

Sent from my iPad

> On 3 Jun 2017, at 21:53, philipjohn <notifications@github.com> wrote:
> 
> I think there are always enough people willing to do the job - we're never short of people standing for election, after all
> 
> That's perhaps true at some levels, but definitely not all. The more local you get, the quicker we run out of people. Take a typical parish or district ward of a few thousand people - we already get co-option a lot (three times in the last 6 months near me alone) because of a lack of candidates, and limiting representatives to two terms will only exacerbate that.
> 

We are not talking all levels just parliamentary, so that issue does not arise.

> For offices such as mayors, prime ministers or presidents a term limit could be a good way to prevent those offices being held onto by the same person if the controlling party is able to keep them there despite public dissatisfaction. Term limits still address the symptom though, and not the cause, which is ineffective mechanisms for recall or removing an elected representative.
> 
Again we are talking parliamentary representation not mayors, we do not elect prime ministers and we don't have presidents, so not sure what you objection is here.
> We should instead make sure that elected representatives are not able to cling on to power, or change the rules in their favour whilst in power (see Putin/Morsi) rather than arbitrary limits that may end service of a good representative who has clear public support.
> 

Sorry but your examples do not equate to representative elections at all, they are basically dictators.

I find your objections and your examples rather off the point and am surprised you voted before any discussion.

The fact is most seats in our elections are safe and that is because party committees put up standing candidates and we have to break that system if we are to move forward to a more representative democracry.

Time limits are the only workable solution IMHO.

philipjohn:
> We are not talking all levels just parliamentary, so that issue does not arise.

Ah - the proposal says "All elected representatives" and thus would apply at all levels. If we can tighten it up a bit, I think we could get to something really good. For example, limiting to mayors, council leaders, prime ministers, first ministers, PCCs.

> we do not elect prime ministers and we don't have presidents

Yep, that's right, and so if the problem we're trying to solve is a single person holding such a position for a long time without  being explicitly placed there by there by the electorate, perhaps the solution should be to have an elected president instead of party MPs effectively choosing the PM.

> The fact is most seats in our elections are safe and that is because party committees put up standing candidates and we have to break that system if we are to move forward to a more representative democracry.

Yes, that is a problem, but that's because we have First Past The Post. Stopping people even standing for election under that system won't stop that system being terrible. Removing FPTP will solve that problem.

Floppy:
Oh, I have an idea @philipjohn, based on what you said. How about if it was all "full-time" positions? i.e. the ones that are paid as if they are full time? That would exclude most councillors but would certainly mean that politics can't be a "career". And if you ban second jobs at the same time (which we do) it should cover it correctly. Would that work?

Motobiman:
Answers below.

Sent from my iPad

> On 4 Jun 2017, at 20:41, philipjohn <notifications@github.com> wrote:
> 
> We are not talking all levels just parliamentary, so that issue does not arise.
> 
> Ah - the proposal says "All elected representatives" and thus would apply at all levels. If we can tighten it up a bit, I think we could get to something really good. For example, limiting to mayors, council leaders, prime ministers, first ministers, PCCs.
> 

We are talking government here not local elections, elected mayors or PCCa and we don't elect prime ministers or council leaders.

That would be too big a single step. IMHO

Ok so say for parliamentary elections only.

> we do not elect prime ministers and we don't have presidents
> 
> Yep, that's right, and so if the problem we're trying to solve is a single person holding such a position for a long time without being explicitly placed there by there by the electorate, perhaps the solution should be to have an elected president instead of party MPs effectively choosing the PM.
> 

That's not the problem we are trying to solve because we don't elect prime ministers and the terms president does not apply in GB.

That's a different argument altogether unconnected to time limits on MPs standing in post.
> The fact is most seats in our elections are safe and that is because party committees put up standing candidates and we have to break that system if we are to move forward to a more representative democracry.
> 
> Yes, that is a problem, but that's because we have First Past The Post. Stopping people even standing for election under that system won't stop that system being terrible. Removing FPTP will solve that problem.
> 

Again that's another problem unconnected to this one.

philipjohn:
That's a neat idea @Floppy - I'd like that, but I'd like to exclude MPs who aren't government ministers. There are a bunch of MPs who have served for a long time because they've been really good representatives, and I don't think we should force them out. I appreciate there are also plenty who have been in for a long time who are *not* good representatives, but that problem can be dealt with by changing the electoral system.

Motobiman:
But we are not talking about changing the electoral system but introducing limits on how many parliamentary sessions and/or years an MP can stand for and when he can stand for reelection.

Changing the candidate selection process and moving to a more democratic voting system than FPtP is a different argument.

Surely the admission you make that plenty of MPs are not very good is sufficient reason to place the limit, or are you happy to only have a minority of MPs doing a good job.

I'd rather have the majority, personally.

philipjohn:
The problem is this;

It is possible for poor quality representatives to remain in office for very many years.

This proposal is addressing that by limiting terms in office. My problem is that while it solves the problem, it would also force many good representatives to stop serving their constituents - a situation which would not be good for them.

The root cause of the problem, as stated above, is our electoral system.

So no, they are not separate issues. The point I'm making is that I accept there is a problem, but do not believe this proposal is a solution because it doesn't address the root cause and instead would have a clear negative impact.

I think there is *some* justification for term limits for some offices. Coupled with a better electoral system would address the problem.

Motobiman:
Comments below.

Sent from my iPad

> On 4 Jun 2017, at 21:32, philipjohn <notifications@github.com> wrote:
> 
> The problem is this;
> 
> It is possible for poor quality representatives to remain in office for very many years.
> 

Agreed.
> This proposal is addressing that by limiting terms in office. My problem is that while it solves the problem, it would also force many good representatives to stop serving their constituents - a situation which would not be good for them.
> 
You have moved from some good reps to many.
The end justifies the means.
> The root cause of the problem, as stated above, is our electoral system.
> 

The issues with the FPtP system is not the same is that of sitting MPs IMHO
> So no, they are not separate issues. The point I'm making is that I accept there is a problem, but do not believe this proposal is a solution because it doesn't address the root cause and instead would have a clear negative impact.
> 
Yes they are separate issues and curing one does not cure the other.
> I think there is some justification for term limits for some offices. Coupled with a better electoral system would address the problem.
> 

OK so now we agree their is justification for term limits.

And we also agree the electoral system need to change too.

Can I count in your voting for the proposal then?

Floppy:
@motobiman, I think @philipjohn would be happier if we looked at it with the "full-time" specifier I proposed - what do you think? You can edit the text to add that, if you like it.

Motobiman:
The fact is we already pay MPs a full time salary and then they get a ministerial position and apparently are so clever they can do two full time jobs.

Some people can't even get one job at a minimum wage or work full time in the NHS and will have to use food banks.

Go figure it.

I maintain my position that MPs should be limited to two terms in parliament and not be able to stand for ten years after the end of the second term.

If this means a few good ones are kicked out with all the dross, so be it.

Autumn-Leah:
Career politics isn't necessarily a bad thing if they're doing a good job. If I've trained for 20 years in politics, economics, and climate science, gotten degrees, and worked in the industry, and then get consistently re-elected by my constituents to fight the good fight re: both climate change and representing them, why should I be blocked from doing my job a bit longer.

The issue with someone holding office for a long time is:

- The representative thinking they can get away with anything, solvable by allowing constituents to trigger a by-election, which is a proposal I'm just about to write, thanks for the inspiration :D

- The representative being stuck in their ways and refusing to look at new info or consider differing perspectives, solvable via the ballot box, the above no confidence idea, and the transparency measures already set out in the Something New manifesto

- The representative getting influenced by other jobs or where their political funding comes from, which is already dealt with under current Something New policy

- The representative *actually* getting away with doing whatever they want to because people vote for the party, solvable via Something New's policy to introduce STV

Vote: ❎

Autumn-Leah:
If you can provide me with another reason I should object to career politicians I'm all ears

Motobiman:
Because career politicians often have no life experience outside of the academia that knows everything in theory and nothing in practice, are selected as candidates by closed shop, unelected party committees and as they are largely unemployable outside politics, will say anything and do anything to keep their jobs.

Oh and we have to stop parachuting as well. . . .  .

Motobiman:
And you just gave several more reasons all valid.

Its party politics that is the elephant in the room.

Such a system that puts party above country, permits candidate selection in secret by unelected bodies and is open to unelected pressure groups needs to be rehashed.

Membership of a political party should be grounds for refusing a candidates offer to stand, it is for some organisations, selected by the same parties that cause the issues.

Judge and jury in their own cause then.

Autumn-Leah:
> Because career politicians often have no life experience outside of the academia

Then make a proposal that mandates candidates publish their CV for the electorate, with no official employment (that is, work that isn't volunteer or illegal, since one isn't necessary if they don't want to include it and the other should appear on their criminal record) left out. Then make it an offence to knowingly lie in said CV.

> that knows everything in theory and nothing in practice, are selected as candidates by closed shop, unelected party committees

This has nothing to do with their viability as MPs, if someone from a party wants to run in that constituency, they can, if they don't they don't, if they stay or are kicked from their party for doing so or otherwise is a matter for them, since they chose to join the org. These rules only matter in elections if the people who'd otherwise run don't want to leave the org, and that's their choice, so it's not really a downside.

> and as they are largely unemployable outside politics, will say anything and do anything to keep their jobs.

Then they'll be seen as liars and removed from power. It's not the existence of parties, or politicians seeking a career in parliament, that causes MPs to be unskilled, and that's on the presumption that your statement about these MPs being unemployable is actually the case.

Motobiman:
All spoken like a true politician.

Let's call a spade a spade.

Everything politicians, touch ultimately fails.

I am 69 years old and have watched this country go from being the gold standard in running and making anything to a sad joke and all because our political system is designed for and by politicians.

Look at just one thing, taxation.

Let's see you defend the abortion that is the tax system.

Motobiman:
Lots of words saying very little.

KISS

Just reduce the number of seats to 100, mandate time limits on standing, make membership of a party grounds for refusing a candidates application to stand and for deselection if elected, permit one vote per session to each MP under the whip, remove enabling acts and remove FPTP.

There would be no professional politicians because there is no career.

That's joined up thinking, a process apparently entirely missing from modern politics degrees.

Autumn-Leah:
> Look at just one thing, taxation. Let's see you defend the abortion that is the tax system.

Firstly, I see no reason why I should defend a tax system created by a government that I fundamentally disagree with, and that isn't caused by career politics but by the special interests that fund them (I'm talking about the Tories if you haven't caught on).

Secondly, I don't see how your obvious opposition to career politics has anything to do with me needing to defend the tax system for some unstated reason.

>  make membership of a party grounds for refusing a candidates application to stand

So you'd leave parliament to the rich and privileged? If everyone's indie then how do they get funding to run a campaign? You'll have the people rich enough to get TV spots and the people who aren't, and unlike our current system where the small parties get left in the dirt, you'll have a worse system where everyone running other than rich people get left in the dirt.

Motobiman:
Glad you don't seek to defend.

Why would banning party membership open an already open door?

The fact is the world is changing and politics is not changing with it.

This result of the last election proves it.

I really believe we have seen the start of a seed change in voting habits as the old tribal left/right party voters die off.

The young are not so easily fooled by money and social media will level out yuckspeak and dogmatic politics.

Traditional politics and politicians are yesterday and if they will be made to change by circumstances.

We cannot continue with divisive and confrontational politics.

RobertWootton:
It seems to me from this debate that a Constitutional Convention Is called for. An organisation has recently been formed called Independent Constitutionalists UK; website is ICUK.life. My personal view is that all people elected to Public Office should be paid a wage and classed as a Public Sole Trader and required to publish annual audited accounts just as a PLC has to do. Expenses claimed should be "wholly and necessarily" incurred for the running of their office. Perhaps 10% of their income could be an allowed expense towards the party of which they are a member.

RobertWootton:
To stand for election, a person must already be a committee member or a volunteer with a charitable organisation, e.g. a Food Bank, W.I., Citizen's Advice Bureau, etc. Such a person would have experience of the social problems people face and hopefully be better placed to find solutions.

Motobiman:
It's the party membership that I have a problem with.

Any organisation that seeks to change government policy by putting pressure on elected representatives is fundamentally anti democratic.

Membership of such an organisation should be cause to reject any application for election and or cause for a by election if joining during office.

Our elected representative have to be freed from party control.

RobertWootton:
SomeThing New is a political party. If the convention of the Party Whip was outlawed in parliamentary votes of all parties would not all parties be “democratic”?

RobertWootton:
So the Green Party is undemocratic as is UKIP?

Motobiman:
It goes further than that into local elections too.

All candidates for major parties are selected to stand by unelected committees and in most cases the independent does not get a look in.

Motobiman:
Of course.

Any candidate that is selected to stand by his party is basically corrupting the democratic process so by extension so is the party system.

By the people, for the people, of the people.

What we have is by the party, for the party and by the party, as the past three decades of totally nonsensical, party-political-ethos government, has shown.

PR #580, opened 2017-06-02
improvement to non-discrimination principle

Floppy:
Given recent discussions on this section, which was getting long and unwieldy, I've tried to condense it down into something more readable, without losing any detail. I've taken the list of categories into a footnote, and used wording from the UN Declaration of Human Rights in its place. I've also taken a general definition of discrimination (being about perceived categories) from Wikpedia. I've tried not to make up my own words here, just put others together :)


philipjohn:
Vote: ✅

Floppy:
Thanks @xyleneb, I did hope this would resolve some of your concerns at least :)

PR #586, opened 2017-06-04
Added: Vote for life for overseas voters

Julia1414:


Floppy:
As long as they're citizens, they should get a vote. Absolutely.

Vote: ✅

philipjohn:
Yep, if you are a citizen and have rights under British law, you should have a say in governance.

Vote: ✅

Xyleneb:
I don't think you should get to tell the rest of us how we should live if you were last here in 2002.
I would expect to give up this right myself if I moved.

This part intrigues me though: "whose rights abroad are dependent on their citizenship".
What is meant by this?

Julia1414:
Hi @floppy and @philipjohn - thanks for supporting this.

@xyleneb to add more context, and to address your remark:   Being able to participate in democracy is a fundamental human right, and should not be taken away because of your personal choices, such as living in a difference country.  I have a friend in his 20's, moved overseas five years ago, never registered in the UK which means he can't vote as an overseas citizen.  It's not always the 15 year rule that is the block here.  Either way, my argument about it being a fundamental right stands.

I do however believe that it's pretty pointless how the current system works - which is why I also added the introduction of MP's who specifically represent overseas citizens as to me they go hand in hand. Overseas voters have a specific set of issues, and they are not the same as a resident voter.  

As for your question - my rights to live, work, travel in the Netherlands, within the EU and beyond is based on being an EU citizen, which comes from being a British citizen.  There is a codependency.  With Brexit looming, it's an issues high over a lot of overseas voters minds.

PR #587, opened 2017-06-04
Added:  MPs for overseas constituencies

Julia1414:


Floppy:
Vote: ✅

Floppy:
In practice the boundaries would be decided by a boundary commission in the same way as existing constituencies, but I don't think this policy overrides that implementation detail.

philipjohn:
Would this add MPs to the commons? I'm wary of supporting that, because I think we have more than enough :)

How about having expats' votes be counted towards the constituency they resided in before they emigrated instead?

Floppy:
@philipjohn that's what currently happens. The MPs for this should be included in the reduced total, so no, I wouldn't expect it to add more. Just require boundary changes to adapt to the right figure.

philipjohn:
Oh, I was under the impression it wasn't the case already. I'm wondering how we'd deal with boundaries, especially as we're changing to STV, but that's implementation detail. Ship it!

Vote: ✅

Floppy:
Yep, boundaries are down to non-governmental commissions anyway, so we don't need to solve the problem, jsut say that it should be solved! :shipit:

Julia1414:
Thanks @philipjohn & @floppy  

This comes down to how you answer the west lothian question :-)

PR #588, opened 2017-06-04
Triple lock for education funding

Floppy:
If anything should be triple locked, it's schools.

### DIFF
--- a/education.md
+++ b/education.md
@@ -6,7 +6,7 @@ We believe that education is the keystone that makes all other longer-term strat
 
 ## Funding
 
-Our schools must be funded properly. We will halt any planned school funding cuts, and introduce a so-called triple lock for investment in state schools, where their annual funding must increase at whichever is greater of 
+Our schools must be funded properly. We will halt any planned school funding cuts, and introduce a so-called triple lock for investment in state schools, where their annual funding must increase at least at whichever is greater of:
 
 * the rate of inflation
 * average wages
--- a/education.md
+++ b/education.md
@@ -6,6 +6,12 @@ We believe that education is the keystone that makes all other longer-term strat
 
 ## Funding
 
+Our schools must be funded properly. We will halt any planned school funding cuts, and introduce a so-called triple lock for investment in state schools, where their annual funding must increase at whichever is greater of 
+
+* the rate of inflation
+* average wages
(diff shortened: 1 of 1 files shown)


philipjohn:
Fuck yes.

Vote: ✅

Floppy:
I missed out "at least", making it seem like it couldn't rise faster than those things if we wanted to.

philipjohn:
Vote: ✅

Xyleneb:
I'll probably vote yes either way, but would you two agree to earmark the surplus for school expansion and additional staff, rather than mere wage increases?

philipjohn:
I'd counter that *where* the money goes should be a decision for schools to make, not government ministers ;)

jt-nti:
"If anything should be triple locked, it's schools." +1
Even comes with a ready made, 'education, education, education' tag line for marketing purposes!
(I also agree with @philipjohn that it should probably be up to schools to make spending decisions.)

Xyleneb:
The secretary of state for education has one job. Being an administrator, not an educator.
He or she decides where funds get allocated. Yeah, maybe he shouldn't be that prescriptive about it. But we already have some of the highest paid teachers. We out-rank the rest of the EU on this. If you devolve this power to the education ministry too, then at what point does the public have a say on how their taxes get allocated?

Floppy:
I'm not sure, honestly. I'd be very wary of saying how schools should spend their money, myself. I'd want someone with more direct experience to make that call. But I don't see why money wouldn't be used for wage increases, if necessary. Schools would take that decision in the same as any other organisation would.

Also, I have no idea who those wages are currently set by.

Xyleneb:
I don't mind how much you increase their budget by. I'm happy to bankrupt all of the other government services for education. You can't have state security, or health or businesses if your people are stupid.

Pledge to spend evermore here and I'll likely be fine with it.

Vote: :white_check_mark:

I'd just be worried that:
1. The electorate won't be fine with it.
2. They'll suspect that we didn't budget properly (leading to public debt).
3. Throwing money at the problem won't fix it. You've got dilapidated buildings, lack of staff, lack of schools. And it won't be solved simply by making sure the pay is good.

Insist that the money will be spent intelligently and you go some step towards reassuring against these things.

PR #591, opened 2017-06-04
Triple lock on NHS funding

Floppy:


philipjohn:
Vote: ✅

Floppy:
I missed out "at least", making it seem like it couldn't rise faster than those things if we wanted to.

philipjohn:
Vote: ✅

Xyleneb:

I think that we should pledge to maintain the NHS budget in real terms.
I don't think that we should offer the NHS a budgetary increase (in times of low inflation, by 2.5%).

We can't sustain an NHS that continues to consume so much money.

Floppy:
I see it more as a commitment to continuing investment in health, but yes, at some point presumably you'd have to remove the escalator. I avoided setting a specific percentage here partly for that reason.

Xyleneb:
Ahh, right. Well if wages go up 2%, then should the whole NHS budget go up 2%?

Autumn-Leah:
Vote: ✅

Autumn-Leah:
Actually, no, the wording on this is pretty awful.

Vote: ❎

Floppy:
Point taken re wages and other things. Triple lock isn't relevant here perhaps, this was a bit of reaction to all the talk of pension triple locks during the campaign. Let's scale this back a bit:

I'd like to find a way to improve this so that funding at least has to stay the same in real terms, i.e. track inflation. But of course the budget could go down for good reasons, or go up. I want to say that *proper real budget changes aside*, NHS budgets should increase with inflation. Any thoughts?

PR #592, opened 2017-06-04
Remove the fixed-percentage escalator on state pensions

Floppy:
No triple lock, just a double on inflation and wages.


philipjohn:
Vote: ✅

Autumn-Leah:
*sips policy*

needs more words

Seriously though, you've gotta explain what exactly you mean, because this can be really easily misinterpreted.

Vote: ❎

PR #594, opened 2017-06-05
Added new section about means testing Christmas Bonus

JackCox83:


Floppy:
I agree with means testing it, though I'd be concerned over the extra cost of doing so outweighing the saving, but presumably if it rolls into other things with universal credit then that should be easy.

On a more general note though, why do we even *have* a £10 xmas bonus?

Vote: ✅

JackCox83:
That's it, currently most people with a state pension or claim certain benefits get a tax free £10 the week before Christmas. It could be so much more effective I feel used to combat a certain issue. These are the current benefits that are entitled to it:
Back to Education Allowance (people coming from jobseeker's payments must have been on their payment and/or BTEA for 15 months)
Back to Work Allowance
Back to Work Enterprise Allowance
Back to Work Family Dividend
Basic Supplementary Welfare Allowance (long-term only - 15 months)
Blind Pension
Carer's Allowance and Benefit (including half-rate Carer's Allowance)
Community Employment
Deserted Wife's Allowance and Benefit
Disability Allowance
Disablement Pension and Widow's, Widower's or Surviving Civil Partner's Pension under the Occupational Injuries Scheme
Domiciliary Care Allowance
Farm Assist
Gateway
Guardian's Payment (Contributory) and (Non-Contributory)
Invalidity Pension
Job Initiative
JobBridge (people coming from jobseeker's payments must have been on their payment and/or JobBridge for 15 months)
Jobseeker's Allowance (long-term only - 15 months (over 390 days of unemployment))
Jobseeker's Transitional payment
Magdalen Laundry Payments
One-Parent Family Payment
Partial Capacity Benefit
Pre-Retirement Allowance
Rural Social Scheme
State Pension (Contributory) and (Non-Contributory)
Tús
VTOS (people coming from jobseeker's payments must have been on their payment and/or VTOS for 15 months)
Widow, Widower's or Surviving Civil Partner's Pension (Contributory) and (Non-Contributory)

dartiss:
I disagree with the plan to means test it. Indeed, I've submitted a general proposal about means testing as I believe it's an unfair way to allocate resources, particularly for those people just outside the bounds of testing.

To give you an example, there's been lots of discussions recently about benefits to the elderly and those receiving cold weather payments. The thing is, it doesn't take many of these tests for those people just above whatever cap there is to end up worse off than those just below.

In all cases, at least, I believe more thought should be made of finding another solution, other than means testing. I accept that, in some cases, there may not be a fair alternative but I think those instances should be limited.

So, back to this one, yes, why do we have it at all? What's it achieving? If it's to help those most in need on the run up to Christmas, why not have it as an uplift in benefits around this time?

Floppy:
Thanks @dartiss, and thanks for opening the other discussion on means testing in general, it's exactly how this should work :)

Autumn-Leah:
Vote: ❎

philipjohn:
I'm gonna say no, because I'm questioning why there is even a bonus in the first place. I'm not even a fan of bonuses in general, to be honest.

Vote: ❎

PR #595, opened 2017-06-06
Suggestions for the future of the BBC

dartiss:
General taxation for the licence fee, non-government control but tight budgets, keeping close to the original aims of the BBC and providing an emphasis on unique and quality programming.

### DIFF
--- a/culture.md
+++ b/culture.md
@@ -21,3 +21,11 @@ Encourage the Advertising Standards Authority to tighten regulations around the 
 ## Media Retractions
 
 Media outlet retractions should mirror the original article. This covers font size, word count, page space and positioning within the document. If a media outlet does not follow this it will receive a fine from Independent Press Standards Organisation (IPSO).
+
+## The BBC
+
+The TV licence should be scrapped in favour of general taxation to provide the budgetary requirements of the organisation.
+
+The budget of the BBC, plus general management, should be kept independant of Government, so there can be no interference.
+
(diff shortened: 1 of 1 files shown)


Floppy:
In general I'm OK with this, though I think that the diversity of the BBC's output is what lets it fulfil that public service role. Could you elaborate a litle on what the reduction in channels and radio stations would mean? I wouldn't want to lose CBeebies, for instance, even though there are many commercial kids' TV channels.

Xyleneb:
It's an insulting waste of public money to offer subsidies in whole to the media. South Korea does it and "Gangnam Style" is the best that they've produced by it.

I don't think I'd ever vote for anything short of abolition.

Autumn-Leah:
I'd vote for it but I don't agree with the proposed solution.

> By ensuring emphasis is given to the original aims of the BBC (that is, to provide quality programmes of the like that you wouldn't get on commercial channels) and of bringing forward new talent, budgets should be kept to a minimum, through a reduction in TV channels and radio stations and by less emphasis on established but expensive TV personalities.

should be

> Emphasis is given to the original aims of the BBC (that is, to provide quality programmes of the like that you wouldn't get on commercial channels) and of bringing forward new talent, by less emphasis on established but expensive TV personalities.

Vote: ❎

philipjohn:
I like this in general. The third paragraph does kinda contradict with the second paragraph though I think - it can't be independent if we're saying we'll change what it does :) Would you be open to slimming it down to just the first two?

Vote: ✅

PR #596, opened 2017-06-06
Added section on means testing

dartiss:
Means testing is an unfair system if allocation of resources, particularly for those just outside of the boundaries of the means test.


Floppy:
Can you provide a source for the "recognised as an unfair method" part? I'm split on this as I would like to see a non-means-tested basic income, for example, but in the current state of our economy, means testing allows focus on those who really need it. Perhaps it's a transition thing?

PR #598, opened 2017-06-11
Added a new section on flooding

simonneb:
New section on flooding says that no new builds on floodplains will be allowed without the new builds providing adequate protection to themselves and other existing properties in the area

### DIFF
--- a/environment.md
+++ b/environment.md
@@ -34,6 +34,9 @@ We will campaign within the EU for reform of the [EU Emissions Trading Scheme](h
 
 We will push for increased climate action and leadership within all international organisations, such as the EU and UN.
 
+## Flooding
+We will ban building on floodplains unless the properties are designed with integrated flood protection systems that offer improved protection to not just the new builds, but existing properties in the area.
+
 ## Home Insulation
 
 It's widely acknowledged that insulation (e.g. cavity wall or loft) is the biggest, quickest way of improving the energy efficiency of the nation's housing stock.
--- a/environment.md
+++ b/environment.md
@@ -37,7 +37,7 @@ We will push for increased climate action and leadership within all internationa
 ## Flooding
 We will ban building on floodplains unless the properties are designed with integrated flood protection systems that offer improved protection to not just the new builds, but existing properties in the area.
 
-We will ensure that insurance payouts due to flood damage will be used for changes to the prpoerty to ensure any future flooding has a negligable impact.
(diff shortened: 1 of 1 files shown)


Floppy:
Great stuff, and a good new section overall - I have things I'd add later on :)

Vote: ✅

Xyleneb:
If you told me that in order to build my floating house, I had to make yours float too...
I would tell you to get lost; I'm not rebuilding your house for you.

If you told me that in order to insure your house I had to prop up it's flood resistance to the point of diminishing returns; again I would pull out of that market.

The part of point 1 that's usable should already be covered under planning permission. Unless you say how you're going to address it (where these builds slip through the 'planning permission net') you ought to discard it.

Point 2 has merit to it, if you detail how you will encourage insurers or how you will pool that money into flood defences.

The regulations as written will worsen the problem of lack of new (resistant) homes, and lack of affordable insurance in these towns.

Please work on it

Autumn-Leah:
I'd have voted it through if you'd've not added the littering amendment, I'm not one for littering but £20 per piece is pretty damn steep, especially given this country and it's history of police corruption. I don't want to be left £40 short because I dropped 2 things by accident and a nearby cop doesn't like me enough just to ask that I pick it up.

Vote: ❎

Xyleneb:
>I'd have voted it through if you'd've not added the littering amendment

The policy results in:
fewer new homes built
higher insurance costs for old homes
a prohibition on spending your insurance money on the goods you lost (that you might well need to live)

You can call it well-meaning all day, but the guy needs to work on the end results.

simonneb:
I have removed the litter section and will re-add this to a separate pull request later - can you relook at it please @Autumn-Leah ?
@Xyleneb you raise some great points. the main point i want to make is this: if your home floods and you get a payout, a proportion (by law) of that payment should go into ensuring that the property is fitted with flood mitigation measures. this doesn't have to be a wall or other hard defence, but instead fitting the internals of the property with things like hard floors, water resistant plastering, installing sockets higher up the wall. that way, any future flooding to the property will have less of an impact both on the property and individual. it would also make it more attractive to future buyers, and also insurers.

philipjohn:
The first paragraph makes sense, and would likely be implemented through changes to planning law.

The second paragraph isn't ideal though - insurance payouts are there to enable households to replace damaged possessions and reparations to the building itself. The reality is probably that flood protection is a much much wider issue than just the single household and it's going to harm the householder to put that cost on them.

Vote: 🤐

PR #600, opened 2017-06-12
Small Crime & Justice Repeal

Xyleneb:
Mild grammar changes.
Verbage on the psychoactive substances act is out.
Points regarding sexual health and abortions are out too.

I'm not sure how much I agree with the policies themselves, but you can assume that these things will happen given your stated aims.

### DIFF
--- a/crime.md
+++ b/crime.md
@@ -4,7 +4,7 @@ title: "Crime &amp; Justice"
 
 ## Capital Punishment
 
-We are opposed to the death penalty for all crimes, and would vote against any attempt to reintroduce it. We will also lobby internationally to stop the death penalty in other countries.
+We are opposed to the death penalty for all crimes and would vote against any attempt to reintroduce it. We will also lobby internationally to stop the death penalty in other countries.
 
 ## Removal of British Citizenship for Criminal Acts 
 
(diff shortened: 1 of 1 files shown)


Floppy:
As it's not just wording, but a couple of removals of policy points as well, can I ask you to resubmit the policy edits as separate proposals (the psychoactive substances and sex work paragraphs)? Then we can discuss those separately.

Xyleneb:
>As it's not just wording, but a couple of removals of policy points as well, can I ask you to resubmit the policy edits as separate proposals (the psychoactive substances and sex work paragraphs)? Then we can discuss those separately.

If the discussion gets confusing, yeah. I didn't think it'd be too difficult.

Autumn-Leah:
> The use of sexual protection would be mandatory at these establishments, and abortions would be provided without charge, as it is under the NHS.

It can't just be naturally assumed that these will be included under the current legal system, if it could I'd have never put it in the policy in the first place, and I don't see the value in removing it.

I don't yet know my stance on the psychoactive drugs policy, but given what you've take off I can't let this pass, equal treatment can't just be assumed, it has to be enshrined, or those affected risk losing it.

Vote: ❎

Xyleneb:
>It can't just be naturally assumed that these will be included under the current legal system, if it could I'd have never put it in the policy in the first place, and I don't see the value in removing it.

Now I know I'd rather repeal the whole policy. But;

I don't know of any developed nation which has both legalised brothels _and_ neglected to offer sexual health services at these brothels. Nobody wants a repeat of the AIDS scare in the 80's.

"Free abortions for everyone" is already technically and pragmatically true. Specifically bringing it up here, either sounds like a cavalier attitude to use of the health services, or that you'll make brothels function as drop-in abortion clinics as well. Either way the public is going to NIMBY it to high heaven.

>I don't yet know my stance on the psychoactive drugs policy, but given what you've take off I can't let this pass, equal treatment can't just be assumed it has to be enshrined, or those affected risk losing it.

If your policy is to legalise the drugs, then specifying the psychoactive drugs act after that policy is redundant. You're legalising psychoactive drugs. That act is void.

I'm not seeing the propensity for unequal treatment that you're seeing here. If you're going to offer legal brothels, you're going to offer condoms. It's not a mad assumption. It'd be mad to operate one without it.

PR #601, opened 2017-06-12
War on Weapons

Xyleneb:
The ban on old ladies carrying anything to defend themselves with is a joke.
The rest of this stuff is so arbitrary and ineffective that I doubt if it ever saved anybody from being victimised. 

How many hand-claw stabbings (either pre or post-1997) have you seen on your block?

### DIFF
--- a/crime.md
+++ b/crime.md
@@ -44,6 +44,14 @@ Sex work is to be completely legalised (barring sex work for those under the age
 
 If not previously established, a 24/7 freephone number will be formally established to report abuse specifically of sex workers and will add to the myriad of techniques currently utilised by law enforcement to avoid this happening. This will contribute to the ongoing battle against human trafficking and abuse of sex workers.
 
+## War on Weapons
+
+We will recognise the god given (or natural law) right to proportionate self-defense and legalise the carriage of non-lethal weaponry, with an exception of within government property.
+
+We will repeal legislation that prohibits the marketing and sale of weapons online, citing the ineffective result of prohibition and the disadvantage placed upon UK-based retailers that operate in an online capacity.
+
+We will legalise the sale and importation of weapons that were arbitrarily prohibited in previous Criminal Justice Acts, including: nunchucks, butterfly knives, samurai swords, hand claws, shuriken, sickle-cords, etc.
+
 ## Police
 
 We will end the politicisation of police forces that took place with the introduction of Police and Crime Commissioners. Instead, Police Authorities will be re-introduced, comprised of representatives of the police force, local elected representatives, local magistrates and members of the public.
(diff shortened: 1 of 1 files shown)


Floppy:
This doesn't sit right with me and I'll need some serious convincing. Honestly, I don't think people should be carrying weapons around at all. I'm definitely going to need some evidence on how weapon restrictions don't keep people any safer before considering an upvote on policy to remove them. 

Vote: ❎

Xyleneb:
First of all, I think that you're already aware that we've banned everything. You name it, we've banned it. What I think you want to know is; under what metric has that ban worked?

Availability;-
Pepper spray has been banned since 1968. Technically, the ban covers every spray bottle as an offense under the firearms act. Yes, _every_ spray bottle so long as it's "noxious". Is it less available? [Not really.](http://www.ebay.co.uk/itm/PEPPER-SPRAY-PEPPERSPRAY-DEFENSE-3x-SPRAY-DE-PIMIENTA-Fast-shipping-to-uk-eu-/263026816505?hash=item3d3d9eb5f9:g:OZ4AAOSwvflZOR8A) [The independent](http://www.independent.co.uk/life-style/gadgets-and-tech/news/amazon-illegal-weapons-for-sale-a6775526.html) seems to think they're available.

Manufacture;-
Ground chilli and surgical spirit. In a spray bottle. Call it hot sauce, call it a dangerous firearm... or let a jury interpret it for you.

Marketing;-
Yes, marketing is banned too. The words "zombie", "military", "tactical" and "special ops" are UK no-nos. In someone's head in an office somewhere, this stopped a crime from happening.

UK retailers can be threatened or fined for not keeping ahead of all of this. German retailers - not so much.

Safety;-
The point that you're going to pull this policy on. Well, you'd think that [knifecrimes.org](http://www.knifecrimes.org/experts-view.html) would be openly partisan on a topic like this, but even they provide a nice summary of how throwing legislation at the problem hasn't made it go away.

I used to hold the same view as them too, namely that I would be willing to inconvenience a hell of a lot of people to not have to find out what it feels like to get stabbed in the spleen. "Throw them in jail", "make them carry a license" etc etc. But in the end, I remember a conversation I once had that went along the lines of; "Would you be content to live in a zoo where you are safe but not free?"

The answer to that, is that you can put no price on your liberty.

Regulations that are not demonstrably a necessity should be abolished.
The burden of proof should fall to those who claim that these laws, are those that we can't live without.

philipjohn:
Likewise, I'm going to need to see some evidence that;

1) the current legislation does not adequately prevent violent crime
2) there are other ways to deal with the problem that work elsewhere in the world
3) that those other ways could feasibly work here

Vote: ❎

PR #602, opened 2017-06-13
No party leader can stand as an MP

simonneb:
Allows party leaders to focus on the role of leading their party and, if successfully elected, the role of PM. Constituents get a proper representative in the House of Commons also.

### DIFF
--- a/elections.md
+++ b/elections.md
@@ -38,6 +38,8 @@ Prospective candidates must have a good connection with the constituency or ward
 
 We will end the practice of the speaker's constituents being denied representation and a lack of choice in elections. The positions of speaker and deputy speaker will not be filled by MPs, they will be jobs advertised, filled and salaried just like any other.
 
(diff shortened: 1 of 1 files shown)


Floppy:
This is in line with existing ideas of separating legislature and executive, and it could start with this. Might need adjusting to fit together with that policy, but I'm in favour of the idea.

Should it be "wants to become" though? Maybe just trigger byelection after leadership contests decide a new leader?

Interestingly, this would prevent me (any many other founders of small parties) ever standing for election again - should there be something to qualify this in major parties only?

Xyleneb:
If your party thinks that they can do both, then why not let them try to do both?
Better representation is what's in it for me apparently, but I kinda like how they're expected to prove their worth by being electable somewhere. I can't see there being much to gain from forbidding it.

Autumn-Leah:
I don't see the issue with a party leader also representing constituents, since the constituents did infact vote for said leader. Them being leader doesn't erase that.

Vote: ❎

simonneb:
all good points,this needs some more finessing. i will retract and come back to it later.

PR #604, opened 2017-06-14
Resubmission: Changes to School Uniform Regulations

Autumn-Leah:


Xyleneb:
Per the 3rd-to-last post I gave in issue #515;

Can you re-word this to account for the abolition of uniforms?

Autumn-Leah:
@Xyleneb I don't see how I can do that cleanly given the wording but if you can edit that in and reply with that version then I've no objections to changing it and trying to pass a version that accounts for that

philipjohn:
Vote: ✅

Floppy:
We'll have to see which merges first and then adjust whichever remains open to fit. Conflicts often happen with longer-lived proposals, that's why we kill them after 90 days :)



Vote: ✅

PR #605, opened 2017-06-14
Resubmission: Adjustment to arms policy (further edits re: genocide)

Autumn-Leah:
Given our government reportedly is aiding genocide in Yemen, and is selling arms to Saudi Arabia, this is a very important policy to add.

Saudi committing genocide: https://www.rt.com/op-edge/380700-yemen-saudi-arabia-genocide/

UN report on how this is against international law: http://www.ohchr.org/EN/NewsEvents/Pages/DisplayNews.aspx?NewsID=21163&LangID=E#sthash.zjA0dW1Q.dpuf

We're selling Saudi Arabia weapons: https://www.theguardian.com/world/2016/sep/16/british-arms-sales-saudi-arabia-airstrikes-yemen-civilian-casualties

Oh Saudi once threatened us with terrorism, we're still selling them weapons: https://www.theguardian.com/world/2008/feb/15/bae.armstrade

### DIFF
--- a/foreign_policy.md
+++ b/foreign_policy.md
@@ -35,7 +35,9 @@ EU citizens currently resident in the UK should have a guaranteed right to remai
 
 ## Arms Trading
 
-Commit to ensuring that no British-built arms are sold to nations that are likely to use those arms against their own people, or to any other agent or third party that may do so.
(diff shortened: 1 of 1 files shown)


Xyleneb:
I think that "and no arms owned (or formerly owned) by the state at any level of governance" is perhaps a worthy distinction, because it includes arms not only built here but imported to here.

The rest of it seems verbose or redundant to me. "The state" is already quantified prior as "U.K.o.B.N.I".
The UN definition of genocide is unenforceably broad, and the way out is to simply deny it.
The extent of "commit to ensuring" remains imprecise and non-committal, despite the aim being otherwise.

The basic ambition of the policy is good, but the reality is poor.
I think with time it can be improved and redeemed.

Vote: :white_check_mark:

philipjohn:
Vote: ✅

PR #606, opened 2017-06-14
Change to non discrimination principle

ghost:
The 2 changes I've made are from the version before the current one.

Previous version: Be treated equally by the law, regardless of gender identity, disability, sex, romantic orientation, sexual orientation, ethnicity, race, perceived ethnicity and/or race, age, religious and/or spiritual belief, occupation (current or previous), status of union membership (including the type of union and what union), or marital status.

Current version changed because of formatting: On the basis of gender identity, disability, sex, romantic orientation, sexual orientation, ethnicity, race, perceived ethnicity and/or race, age, religious and/or spiritual belief, occupation, union membership, marital status, or any other status based on the group, class, or category to which a person or thing is perceived to belong.

The important distinction is in the 2 adjustments:

> status of union membership **(including the type of union and what union)**

and

> occupation **(current or previous)**

Imagine this, you're in court, you used to work for a company that did some shady stuff and that a member of the jury or the judge doesn't like, now it's not very likely, but legally they can now discriminate against you and use that to prosecute you if they make a good enough argument *even if it has nothing to do with the case whatsoever*. I'm not saying it'll happen, but surely it's worrying that under our plan that would be *legal*. It's the same case for the union adjustment.

### DIFF
--- a/index.md
+++ b/index.md
@@ -18,7 +18,7 @@ We believe that individuals should:
 * Be free to act in any manner that does not harm another individual, and does not infringe upon the rights of other individuals.
 * Enjoy a fundamental right to privacy from the state or their agents.
 
-[^discrimination]: On the basis of gender identity, disability, sex, romantic orientation, sexual orientation, ethnicity, race, perceived ethnicity and/or race, age, religious and/or spiritual belief, occupation, union membership, marital status, or any other status based on the group, class, or category to which a person is perceived to belong.
+[^discrimination]: On the basis of gender identity, disability, sex, romantic orientation, sexual orientation, ethnicity, race, perceived ethnicity and/or race, age, religious and/or spiritual belief, occupation (current or previous), union membership (including the type of union and what union), marital status, or any other status based on the group, class, or category to which a person is perceived to belong.
 
 In short, we wish to ensure government remains open and transparent, whilst individuals retain liberty and privacy.
 


Floppy:
Yeah, I dropped these in my refactor edit, but I'll happily accept them back in. Wordiness isn't so much of an issue now it's footnoted.

Vote: ✅

Xyleneb:
You can fill it with writing that can't be taken seriously, but I can't see what good it would accomplish.

>including the type of union and what union

The law's never going to treat my cactus-wife equally. Just because she's prickly. What are you going to do to ensure she's treated the same? Nothing, that's what.

You can take lots and lots of words that mean well but are never enforceable and then relegate them to a footnote to avoid feeling illiberal or to avoid stirring the pot or whatever, but at what point do you expect the footnotes to be functional or otherwise to not be there?

You can tell me each distinction is a meaningful one. By all means make the case. But if you leave any distinction out, _any possibility at all_ then you've treated them unfairly as less meaningful. Concluding the same as "all are equal, but some are more equal than others". So long as you have explicit lists this can't be levelled, and it's a bad idea. All are equal under the law. Period. No "especially if"s. No "except other than"s.

TL;DR my vote here is always gonna be a vote to oppose

Autumn-Leah:
@Xyleneb The entire point of listing various attributes is so that the public is aware of what types of discrimination the party aims to combat. You can't make a law that says "everyone is equal" but you can make a law that says "judges and juries can't discriminate against people in court cases based on a, b, c, d, characteristics".

> All are equal under the law

This should be the case and isn't, which is the point of anti-discrimination protections.

> The law's never going to treat my cactus-wife equally. Just because she's prickly. What are you going to do to ensure she's treated the same? Nothing, that's what.

If said cactus wife contacts the party (or doesn't, the latter point is the key bit) and it's clear that she's being discriminated against then cactusphobic behaviour would be banned in the legal system I guess, just as homophobic and sexist behaviour is.

> All are equal under the law. Period. No "especially if"s. No "except other than"s.

I agree, and this doesn't do that. You have to have criteria to make it actionable though, meaningless statements of "everyone should be equal" protect nobody and aren't actionable.

> You can take lots and lots of words that mean well but are never enforceable and then relegate them to a footnote to avoid feeling illiberal or to avoid stirring the pot or whatever, but at what point do you expect the footnotes to be functional or otherwise to not be there?

These words aren't meant to be unenforceable, they're meant to be a basis from which the party builds any necessary anti-discrimination into its manifesto. I wasn't the one who made it a footnote, I don't care whether it is or isn't though as long as it's there. The footnote serves the function explained in the first sentence in this paragraph.

philipjohn:
Vote: ✅

Floppy:
This isn't the place to get into an argument over the usefulness of the entire list, as this proposal isn't introducing it or removing it. If you want to reopen that discussion again @xyleneb, then feel free to remove the list in a proposal.

Xyleneb:
>This isn't the place to get into an argument over the usefulness of the entire list, as this proposal isn't introducing it or removing it. If you want to reopen that discussion again @Xyleneb, then feel free to remove the list in a proposal.

You won't abide by it, nor by the extensions proposed.

>If said cactus wife contacts the party (or doesn't, the latter point is the key bit) and it's clear that she's being discriminated against then cactusphobic behaviour would be banned in the legal system I guess, just as homophobic and sexist behaviour is.

She can't, she's mute. I can't read her handwriting either.

>I agree, and this doesn't do that. You have to have criteria to make it actionable though, meaningless statements of "everyone should be equal" protect nobody and aren't actionable.

Your list of "things to not discriminate against", while being equally unbacked and unlinked to any particular policy, does not include bald people, epileptics, goths, jocks, hobbyists, or nation states.

>The footnote serves the function explained in the first sentence in this paragraph.

mhmm, so the "public is aware of what types of discrimination the party aims to combat" (and what it doesn't, because those other people they don't matter).

But let's talk specifically about the amendments proposed which you think would be a useful distinction:

>status of union membership (including the type of union and what union)

polygamous sun cult

>occupation (current or previous)

career criminal

Now, will you make the judge not consider these? No? Then how _do_ you intend to abide by it?

It's the mental gymnastics required to say that positive discrimination must be good discrimination that makes the list and it's amendments so irritating. You won't settle it until there is no list, and that won't happen anytime soon because the list is full of this "just trying to help" feel-good liberal drippings that it's difficult to voice an opposition against.

I'll go ahead and submit it but it will only prove the point.

PR #607, opened 2017-06-14
Wording change

Autumn-Leah:
If we don't change the wording then we've just made a statement against all violent sports, surgery, any military or police action that involves having to hurt people, and all of BDSM. I think it's a good idea to change the wording :D


Xyleneb:
Vote: :white_check_mark:

Only opposition I could think of was "I don't like the Carry On films"

Floppy:
Fair point.

Vote: ✅

geeksareforlife:
Vote: ✅

PR #608, opened 2017-06-14
Accountability of elected representatives

Autumn-Leah:
If you've pissed off 30% of people in your constituency to the point that they're willing to sign a petition against you continuing your role, you're probably doing a terrible job, though I'm up for debating the exact numbers.

### DIFF
--- a/democracy.md
+++ b/democracy.md
@@ -135,3 +135,9 @@ To dampen the influence of the City of London in the legislative process, we wil
 
 ##  MPs for overseas constituencies
 Introduce regional representation for overseas citizens into the House of Commons. At least one MP per continent would be a great place to start and would help give a voice to the 8% of British citizens living overseas.
+
+##  Accountability of elected representatives
+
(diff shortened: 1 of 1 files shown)


Xyleneb:
I think it's worth giving a constitutional protection or two a try.

Vote: :white_check_mark:

You can always debate the numbers later. I could see a few other issues with it:
Politicians that do unpopular things might otherwise redeem themselves. Legislation like this is weak unless you list penalties for not adhering. Petitions, particularly electronic ones are susceptible to forgery.

The second one is probably the most critical. The rest are debates you could probably have as the need arose.

simonneb:
I guess there is a risk if the % is too low that you could get caught in a continual cycle of elections with two sides vetoing each others candidates constantly. Con and Lab both got 40% of the GE vote. maybe a threshold above 50% would be better?

:zipper_mouth_face:

philipjohn:
Vote: ✅

Floppy:
I agree, but we already have a statement on recall on the elections page. This is better and has more detail, but we need to match them up. See https://openpolitics.org.uk/manifesto/elections.html#recall-of-representatives

PR #609, opened 2017-06-16
Egalitarian Ammendment to the Non Discrimination Principle

Xyleneb:
Read (what is currently) the last post contained in request #606.
You could also read the rationale in request #566 if you wanted.

### DIFF
--- a/index.md
+++ b/index.md
@@ -12,13 +12,11 @@ We believe that government should be:
 * Ethical: dedicated to creating a sustainable society and economy with progressive moral values.
 * Bottom-up: People should be telling the government what they want from a local level, rather than being told from the top. 
 
-We believe that individuals should:
+We believe that:
 
-* Be treated equally by the law, without discrimination[^discrimination], and be entitled to equal protection of the law.
-* Be free to act in any manner that does not harm another individual, and does not infringe upon the rights of other individuals.
-* Enjoy a fundamental right to privacy from the state or their agents.
-
-[^discrimination]: On the basis of gender identity, disability, sex, romantic orientation, sexual orientation, ethnicity, race, perceived ethnicity and/or race, age, religious and/or spiritual belief, occupation, union membership, marital status, or any other status based on the group, class, or category to which a person is perceived to belong.
+* All are equal under the law and are entitled to equal protection of the law.
+* All are free to act in any manner that does not harm another individual, and does not infringe upon the rights of other individuals.
+* All possess the inalienable right to privacy from the state or their agents.
 
 In short, we wish to ensure government remains open and transparent, whilst individuals retain liberty and privacy.
 


philipjohn:
I think this maintains the aim of the principles while simplifying the statement, helping us avoid having to update the list in the footnote.

Vote: ✅

Xyleneb:
Bloody hell I must have a speech stat +1

Autumn-Leah:
I'm nitpicking a little but "All are equal under the law and are entitled to equal protection of the law" should be "All should be equal under the law and should be entitled to equal protection of the law" because otherwise it sounds like we think everything is going fine and that everything is already equal.

Secondly, why do you want to get rid of the list when it's a list of characteristics that we can take action on? Or do you believe that it's better to have policy that represents these things, because if so I'd rather have the assurance of having that policy already there before it's removed. I see the list as a way of saying "we know there are issues in these areas", whereas saying everyone should be equal doesn't really say much. With a few adjustments I see removing the list as a bit like have the attitude shown in the first 30 seconds of this vid: https://www.youtube.com/watch?v=eaEeEbP16Wg "being united is an ideal, not a plan" - "be united by joining me in doing nothing".

I'm gonna vote against this for the time being but if you can follow this up with ideas to solve the above equality issues that you'd support if put to a vote, or convince me that the list is unnecessary or the like, I'll readily change my vote.

Vote: ❎

Xyleneb:
>I'm nitpicking a little but "All are equal under the law and are entitled to equal protection of the law" should be "All should be equal under the law and should be entitled to equal protection of the law" because otherwise it sounds like we think everything is going fine and that everything is already equal.

There are pros and cons to the wording. "Should be" is correct and accurate, but it's non-committal. It's up there with "may be" and "wouldn't it be nice if". If you're sure, then "you will". Not "it should be" - "it will be".

The problem is lack of policy to back up the promise. Any way you write it is aspirational.

>Secondly, why do you want to get rid of the list when it's a list of characteristics that we can take action on?

Those who didn't make the cut, those who are forgotten or excluded from your pledge to take action; it is not fair to them. You might say "just because I didn't name them doesn't mean they don't matter". But how is it fair to declare your interests for one but not another?

You can try to make your list infinitely long to account for every characteristic or eventuality, but that is impossible. Or you can concede that _in the context of the law_ treating everyone the same is not only equal but fair and equitable.

>I see the list as a way of saying "we know there are issues in these areas"

I could argue that to those who're suffering it's insulting and it's not enough - to promise you'll do something without backing it up.

But my argument is that I don't want to be explicitly mentioned or singled out, whether that aims to be favourable to me or not. If what I hope to be is equal, then I don't want special mention. For some reason I don't fear being forgotten per se, despite being (relatively) young and poor.

>I see removing the list as a bit like have the attitude shown in the first 30 seconds of this vid

"From my non-plan" is not how it's divisive though. If it only divided a "non-plan" it'd be a lot less hostile to deal with.

>...ideas to solve the above equality issues that you'd support if put to a vote...

I had plenty of ideas for the young. Insurance shouldn't discriminate. Employers shouldn't discriminate. Landlords shouldn't discriminate, etc. But these ideas are "divisive". Insurance premiums would climb for everyone else on the news that insurers are legally forbidden from taking age into account. And unless I'm confident of the impact of these policies, I don't tend to submit them.

PR #613, opened 2017-06-18
Introduction of the Offshore Assets Tax

Xyleneb:
Speaks for itself pretty much.

### DIFF
--- a/economy.md
+++ b/economy.md
@@ -30,6 +30,10 @@ As a general rule of thumb, tax policy should focus on simple direct taxation (s
 
 Enable greater decentralisation of taxation by enabling local authorities to set tax rates according to local needs, keep those revenues and spend them according to local need. 
 
+### Offshore Assets Tax
+
+We will implement an Offshore Assets Tax, so that companies operating here but whose head offices remain in other countries such as Ireland, Luxembourg, etc will still pay taxes based upon the assets they hold abroad.
+
 ### Income Tax
 
 The personal income tax allowance will be set at the level of a full time living wage and will rise in line with the living wage. This means that someone working a normal full-time job at the minimum wage would not pay any income tax.


geeksareforlife:
Full disclosure: I work for a company that has its HQ in Ireland and (IMO) pays less tax than it should.

But I'm not sure that taxing assets that they hold abroad is the right answer - wouldn't it be better to do something like taxing their UK turnover, or requiring them to disclose actual UK operating costs, etc?

I completely agree with the aim - getting them to pay more tax, just not sure that this is the right way to do it

Xyleneb:
>But I'm not sure that taxing assets that they hold abroad is the right answer - wouldn't it be better to do something like taxing their UK turnover, or requiring them to disclose actual UK operating costs, etc?

The problem is that they always find a way of saying that they turned it over in luxembourg, and operated it in luxembourg. Once you change it to assets, the worst they can do is deny that they own those assets. Their parent company owns them, or their subsidiary owns them: "Don't tax _me_ boss!"

geeksareforlife:
The problem with taxing assets is that you are effectively stopping international companies (maybe that is your intention?).  If an international company is operating in 5 companies, and they all have a similar 20% tax on assets, then that company will have to pay 100% on their assets.

This is obviously an oversimplified, contrived, example, but it illustrates what I'm worried about.

I'm not entirely sure that what I suggested above is even a good idea!

Companies always find ways around paying tax, I feel that what we should do is make evading tax a criminal offence for companies over a certain size, and invest in actual investigations.

Xyleneb:
>The problem with taxing assets is that you are effectively stopping international companies (maybe that is your intention?). If an international company is operating in 5 companies, and they all have a similar 20% tax on assets, then that company will have to pay 100% on their assets.

Well, you're right in saying that the Offshore Assets Tax would have to be levied on turnover, not profit.
Because otherwise they could carry on claiming "we didn't make a profit this year" as they do _every_ year.

Because it's based on turnover, if your company is the sort to make lots of money, but also spends a lot of money on a really complex and difficult to manufacture product, then you're kind of screwed by the tax system. Because the tax man will come along and take a percentage of the pot away from you, even though your actual profits are low.

I have to say though if your margins are that small then I struggle to see how you're going to own foreign assets of any significant value.

>Companies always find ways around paying tax, I feel that what we should do is make evading tax a criminal offence for companies over a certain size, and invest in actual investigations.

Tax evasion _is_ illegal and investigated. Tax avoidance is legal, but investigations are funded for that as well for the sake of the tax system and for the purposes of monitoring potential evasion. We've been trying this for 20+ years. When a company says "we didn't make a profit in the UK" it's really hard to disprove. You could claim to have spent the money anywhere. You could claim that UK customers visited Switzerland and bought products there. It's really easy to avoid. It's a little harder to say "we didn't turn anything over in the UK" because you've got a presence here, there's a paper trail of receipts all from UK customers buying your product, whether it's in Switzerland or not.

But the problem with a tax on turnover is outlined above.

That's why the Offshore Assets Tax is there. It restricts this problem to multi-national companies with many assets, and it's not so much my idea as a Bernie Sanders idea. The same guy who should have probably been president came up with this one. And he's good. He's good as a legislator anyway, whether he has the makings of a president or whatever.

The rates as per asset ownership can be decided later. Yes, it's vital to get that right. But I think that the principle is sound and that the results will be satisfactory.

PR #615, opened 2017-06-28
Partial Military Policy Revamp

Xyleneb:
Changes:
Security Industry Authority given the responsibility of managing PMSCs. 
Compulsory publication of all PMSC operations within 3 years (as in not just "Defence Inc - Iraq" but "Defence Inc - Helmand Province, Iraq Operational Objective: Etcetera").
Publication of killed/missing lists changed from annual to quarterly.
Return to public ownership of MoD site operation (not associated with runway/aircraft training).
"What is 'autonomous'?" is given a definition ("electrically autonomously discharged").
"Lethal" is removed - to prevent drones from autonomously firing tasers or CS gas.

If anyone has any problem with any of the changes then I can split them up and submit them separately.
Otherwise tell me what you want changed and I'll try to add it.

### DIFF
--- a/military.md
+++ b/military.md
@@ -28,12 +28,6 @@ We will empower the SIA regulatory body to regulate private military and securit
 
 We will push for an international regulatory body and statutes for PMSCs around the world, possibly administered through the UN. The international body could potentially incorporate both the [Montreux Document](https://www.globalpolicy.org/images/pdfs/0917montreuxdocument.pdf) and the [International Code of Conduct for Private Security Service Providers](http://icoca.ch/sites/all/themes/icoca/assets/icoc_english3.pdf).
 
-## Proprietary Test Facilities
-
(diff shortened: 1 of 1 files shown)


Xyleneb:
EDIT: I'll copypaste the "Return to public ownership of MoD site operation (not associated with runway/aircraft training)" policy into here, and submit it separately if this passes. Reading the document was confusing with it in there.

>We will ask all companies contracted to operate Ministry of Defence sites to separate their testing and evaluation businesses from their civil and military training businesses. Contracts for MoD test facilities will then be allowed to expire and all weapons test grounds will return to public ownership, allowing weapons to be assessed based on their merits rather than their profit incentive.

Now the changes should be legible.

Xyleneb:
Note also for later: Putting the SIA as specifically in charge is a potential regression
https://en.wikipedia.org/wiki/Security_Industry_Authority#The_Future_of_the_SIA

PR #616, opened 2017-07-06
Private School's VAT Intro Dropped, Requirement to Admit Poor Students Instead

Xyleneb:
I couldn't think of a better way to write it ~ no "economically disadvantaged" euphemisms were included.

The old policy of introducing VAT wouldn't break up the cronyism and the old guard. I don't think it'd hit their profits either really. People will pay whatever they're asked for, for the right colour tie. It was simply an exercise in the state taking some of the money from it as a profitable business.

Well, I've made it so that to make a profit they have to help the country's poor to do it. oops.
"Complete subsidy" must be adequately enforced too. Otherwise you'll get Eton boys wearing an "even fancier (unsubsidised) tie" so long as it distinguishes them from the great unwashed.

Also, education is not a luxury. Even if you're giving English lessons to the shah of Saudi Arabia it's not a luxury. I don't think they should pay "what ought to be due" for "the sin of teaching English". Even if that's a lucrative profession for them.

### DIFF
--- a/education.md
+++ b/education.md
@@ -54,7 +54,7 @@ Scottish authorities will no longer be legally obligated to appoint unelected re
 
 ## Private Schools
 
-We will remove the charitable status from private schools, and make fees subject to VAT, as private schooling is most certainly a luxury.
+Every paid-for student must in turn completely cover the bursary of a poor student, and in so doing maintain the anonymity of their benefactors. Thusly the ratio of rich to poor students in all private schools, shall be 50:50.
 
 ## Discrimination in Physical Education
 


Floppy:
As a student of a private school, the split would be *hugely* obvious, so any anonymity would last about 20 minutes. I'm not sure about this at the moment, it feels like too much interference to me. Hmmm.

Vote: ❎

Xyleneb:
>As a student of a private school, the split would be hugely obvious, so any anonymity would last about 20 minutes.

The muggle bloods are 1 in every 2. You've got some admin to do to establish who the toffs are. And with half the school being this or that, it doesn't really matter who's what. You can try to pick on a student and use money as a reason to declare him a lesser-human, but then the other half of the school would have a word or two to say about it.

>I'm not sure about this at the moment, it feels like too much interference to me. Hmmm.

There's plenty of left-wing sentiment that wants them abolished. Compared to that, giving them a task to do is being lenient. Also, you know what else (other than school places) would trouble the school's admin? Paying VAT. If you want to get all free-marketeer about it then why don't you make school optional? Hell I'd have preferred an apprenticeship anyway. It's the first time in a while we've had a generation dumber than their parents and it's going to take a lot of gaffer tape (interference) to fix it. 

You could always devolve this kind of authority to a ministry of education later. I've not submitted the policy yet, because I'm not sure about how it's going to work. I might submit it anyway. It could do with some scrutiny.

Autumn-Leah:
My only question is why not both?

Floppy:
Good point indeed, why not both.

PR #619, opened 2017-08-21
Free to choose an alternative economic (operating) system.

RobertWootton:
I have said in my pinned tweet that "How can a fair and just society be built on the foundations of an inherently unfair economic system?" My e-book which was published in 2013 when we were still members of the EU. It was titled "A Common Economic Architecture for the EU". I have since retitled it "Creating Responsible Capitalism". The POD paperback version title is just "A Common Economic System Architecture".
The main components of the architecture are new types of business organisation which would have to be legislated into existence; an eqPLC, (equitable PLC), an sePLC, a social enterpise PLC and a ciPLC, a community enterprise PLC. These businesses would have STATUTORY wage differentials of 20 to 1, 10 to 1 and 5 to 1 respectively. In this way, the argumens of the Left against the Right would be taken out of the debating chamber. To describe a country as Capitalist, Socialist or Communist would depend on the evolving distribution of the different types of businesses within a country's economy.


Autumn-Leah:
this is a message not a proposal

Vote: 🚫

PR #620, opened 2017-08-23
The paradigm that my alternative economic system is based on the science of cybernetics and systems.

RobertWootton:
The policies advocated for putting right the faults of the economy seem to be based on the existing economic system. It does not change the system or empower the citizens to take control of their own lives and homes. I believe mine does.
What follows is an excerpt from the book; "The job of government is to implement the components of the Common Economic System Architecture and the protocols within and amongst them in law. Since this system does not yet exist, it follows that the system will have no effect on any citizen unless and until they voluntarily choose to use the system."

### DIFF
--- a/economy.md
+++ b/economy.md
@@ -1,4 +1,4 @@
----
+Legislate for a new alternative economic framework that is integrated with the tax and benefit system so that poverty, zero hour contracts and the excessive interest charged by loan companies is eliminated. ---
 title: Economy
 ---
 


Autumn-Leah:
You're gonna need to expand on what you mean by this

RobertWootton:
Autumn-Leah; Systems thinking and cybernetics, in my perspective, involves creating Causal Loops (Systems thinking) and negative (error correcting) feedback loops (Cybernetics). For example, the UK government has recently reduced Corporation Tax for PLCs from 20% to 17%. In my alternative economic operating system, an eqPLC which does not as yet exist in UK law would have a statutory ratio between the highest and lowest paid of say 20%. If a Causal loop is then created by linking the rate of Corporation Tax payable to the actual ratio of pay differential, then a 17% or even an 8% Corporation Tax rate would be payable by the new eqPLC. To encourage investment in these new types of PLCs the rate of tax payable on dividends would be zero.

It was reported in the Economist some years ago that it was estimated wealthy Americans had $15 trillion in offshore accounts etc. The assumption that I base this system on is that businesses exist to make a profit. The causal loops are designed to establish fair and just corporate behaviours.

Floppy:
This would actually be best as a discussion in issues. @RobertWootton would you like to add it to https://github.com/openpolitics/manifesto/issues/new, and we can discuss there?

PR #621, opened 2017-08-23
Dual-booting the political economy.

RobertWootton:
This a free upload of my POD book that describes an alternative economic operating system.


Autumn-Leah:
no advertising

Vote: 🚫

Floppy:
Thanks @Autumn-Leah. And while it's not something that should go in as is, thanks @RobertWootton for pointing us to it, and for sending me a hardcopy. I will have a read and see what we could integrate :)

PR #624, opened 2018-01-04
Added a term limit policy to be debated.

ItamarC235:


philipjohn:
I understand the reasons why this kind of thing is proposed, but I don't agree it's a solution. It was detailed in #569 and I offered a bunch of reasons there so I'll just refer to that debate and drop my vote here.

Vote: ❎

Floppy:
I still tend towards the idea, but I will go back and reread the discussion around #569 before voting here.

Vote: 🤐

Floppy:
Crikey, that one went off topic a bit. I'll try to summarise the relevant parts here a bit later on.

Floppy:
@philipjohn so in that one, before it went off topic too badly, we were proposing term limits on full-time positions. This is a tighter set in this one, basically just MPs and the PM. The argument I understand from you is that it's unhelpful to limit the time in office for good MPs (of which there are many) in order to get rid of the bad ones, when electoral reform would make bad MPs easier to unseat, and would be preferable. Would you agree with that summary of your position?

philipjohn:
Spot on!

Xyleneb:
Vote: :zipper_mouth_face:

Even if we concluded that "there _should be_ a limit" I can't really decide _what kind_ of limit. Don't you sort of wish that Obama could have served a third term?

ItamarC235:
It doesn't matter how good a person is, giving one person a great amount of power for a long amount of time can only cause wrong doing, corruption, scandals, you name it, having a universal time limit makes politics fresh, prevents rotten politics and acts as a regulator upon the public officials.
 Not to mention that not having a limit goes both ways, People will be able to vote officials like Trump and Netanyahu over and over again due to political apathy or blind interest if there was no limit, in case of Netanyahu thats the current state.
In my opinion we must have a limit, a universal one.

PR #626, opened 2018-01-15
Replace Vehicle Excise Duty and Fuel Duty: The Trilogy

Xyleneb:
Let me give you an idea of what this does:
1. Stops hydrogen and ethanol alternatives from being mistargeted under 'fossil fuels', so you are allowed these.
2. Permits hybrid cars and tractors.
3. This line's particular eloquence ("maximum continous output of the main driving motor") ensures that:
3A. Multiple driven axles aren't penalised by the tax system (you can have 4x100kW and you will only be charged for one 100kW).
3B. Allows apples to apples comparison of vehicles, as the spec sheet details typical kW output of the motor(s) rather than short-term output that would drain the battery source in under 15 minutes.
3C. Permits temporary under/over-volting, which simply won't be of any concern to the taxman unless you want to re-register your vehicle with the changed values.

It offers a great many incentives to go electric. It takes a system as complicated [as this](https://en.wikipedia.org/wiki/Vehicle_Excise_Duty) and explains it within roughly six paragraphs. I defy you to find a better system anywhere [within here](https://en.wikipedia.org/wiki/Road_tax)...

That said, I'm mainly pleased because it's not a copy-paste policy. It's not just "let's do this". It was hammer-and-chisel effort.

So yeah I'm pleased with it. But your scrutiny, please!

### DIFF
--- a/transport.md
+++ b/transport.md
@@ -8,7 +8,7 @@ No road vehicle powered solely from the use of a petrol or diesel engine may be 
 
 ## Taxi Services
 
-We will incentivise taxi companies to buy taxis with lower or zero carbon emissions via tax reductions on the company's profits, scaled in a carbon brackets system that takes business size, current carbon output, carbon reductions made within the year, and the starting point for having entered the scheme into account.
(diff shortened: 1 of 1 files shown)


Xyleneb:
Also sorry for the number of "commits" I was arguing with the way the github editor does things.

PR #627, opened 2018-04-10
The Republic is a step too far, for now and will turn off older voters

MartiFrench:
The republic idea is a small part of what you are trying to achieve but will turn off older voters who otherwise would join. Step too far, for now. 
Election for ceremonial position are pointless, election fatigue could results which will have low turn out and could result in non serious candidate or divisive figure being elected.
Stick with Monarchy carrying our same role. What would we do with them otherwise?  Monarchy will fade over time, its not important right now.


lucaclarkgutierrez:
:negative_squared_cross_mark:

PR #630, opened 2019-01-05
Added key policies

Periodic-Thunder:


Floppy:
I assume this is a joke submission, though I definitely agree with the cookie clicker proposal. :D

Vote: 🚫

PR #632, opened 2019-01-25
democracy is failing just look at china its moving forward we are in reverse

TerryKing1:
DEMOCRACY -V- BENEVOLENT DICTATORSHIP 
It is arguable that a benevolent dictatorship is better but how can you know if your going to get a benevolent one or total maniac. What we need to do is create a new democracy that opens the door to ALLOW for leaders to deliver their promises and not a system that all we hear is a loud noise of people yelling at each other to ensure that their voice is heard for their own personal view. We all have our view but frankly none of that really matters what is important is REAL LEADERSHIP. When we realise that we will then come together as a nation and all really benefit.


Floppy:
I'm all for improving the way we do leadership, but elected dictators tend to become unelected ones pretty quickly - curbs on power are essential checks and balances. Though even with that we see how little they're worth if the person in charge decides to ignore them (looking at you USA).

Vote: ❎

PR #639, opened 2021-02-09
Added section about nationalised sofware company

JoeRourke123:
Proposed in an effort to reduce delay and cost of outsourced government projects.

### DIFF
--- a/infrastructure.md
+++ b/infrastructure.md
@@ -54,6 +54,13 @@ To be an effective policy, it will also cover any organisation performing one of
 * operating public bus services
 
 
+### Nationalised software development
+The government should oversee a new software development arm, with a technologically adept minister overseeing it. This new nationalised company should be up-to-date with the latest technologies and practices in order to reduce the cost and delays frequently seen in government software projects.
+
+Company employees should be given paid training on new and emerging technologies in order to keep the company in line with current standards.
+
+
+
 ### A Compromise on Nationalisation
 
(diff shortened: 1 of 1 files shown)


Floppy:
How do you see this differing from what the Government Digital Service are doing? https://www.gov.uk/government/organisations/government-digital-service

JoeRourke123:
> How do you see this differing from what the Government Digital Service are doing? https://www.gov.uk/government/organisations/government-digital-service

GDS sets standards for use by contracted companies, I believe. I imagine this office being completely inhouse (no contracting), leading to more cohesiveness amongst projects and allowing for a simpler model for planning/design/distribution of software (as lots of middlemen cut out). Undecided as to whether this would involve the nationalisation of pre-existing companies or starting fresh with establishment of a whole new division. The outcome/reasoning for the division would be to prevent failings like the original Track and Trace app.

Floppy:
GDS does a lot of development in house, and helps other departments to set up their own in-house team. It does also set standards for contractors (I currently am one!) but that's mainly because govt wants to be able to scale up more quickly than it could with everyone in-house. There's definitely a big push towards in-house development and ownership of code. I thought the Track and Trace app was developed in-house by NHSX, but I'm not well-versed in the history, tbh.

There's definitely something here about protecting what GDS does and pushing it further, though. The government has stepped back from the original vision a bit.

PR #640, opened 2021-05-25
Mainly just testing how live this is

eddowding:


Floppy:
Still alive and maintained, though not very active! We're working on generalising the software a bit, more than the actual manifesto itself.

